PR #1, opened 2017-11-29
copy init postman test scripts from IOTech

cloudxxx8:
copy init postman test scripts from IOTech repository


JPWKU:
@cloudxxx8 As you can see there is a check failing that says sign-off is missing.  The EdgeX community requires all committers to sign their commits.  Doing so means the committer attests to this https://developercertificate.org/.  There is no reason to undo the commits, I would squash them together and then ammend the one commit and sign it by adding the `--signoff` flag to your git command.

cloudxxx8:
Thanks for reminder, @JPWKU 
I will sign off my commits in the future.
Please let me know what I can do for this .

JPWKU:
Hi @cloudxxx8 I would squash down these two commits to one...and then do a `git commit --amend --signoff`.  that should do it.  After that you will have to force push to your branch and that will update the PR.

PR #21, opened 2018-05-11
Point blackbox tests at the new core go services

JPWKU:

### DIFF
--- a/bin/env.sh
+++ b/bin/env.sh
@@ -2,13 +2,12 @@
 
 export volume=edgexfoundry/docker-edgex-volume
 export consul=edgexfoundry/docker-core-consul
-export coreMetadata=edgexfoundry/docker-core-metadata
+export coreMetadata=nexus3.edgexfoundry.org:10004/docker-core-metadata-go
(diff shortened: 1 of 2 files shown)


jpwhitemn (changes requested):
Jeremy - we are missing support-notifications

JPWKU:
@jpwhitemn Ah I have extras in there we don't have yet correct?

jpwhitemn:
@JPWKU If we are just testing go core services, then we should just have core data, metadata and command.  Those are the only ones we have blackbox tests for right now.  If you want to have all the services (existing or potential listed) then you need to add notifications.  Wasn't sure based on what you were listing.

JPWKU:
recheck

JPWKU:
recheck

JPWKU:
@jpwhitemn I'm doing stuff locally but I think we should point the blackbox tests at the images in nexus3.edgexfoundr.org staging registry only.  We want to test daily, but we don't push images to dockerhub daily (yet).

JPWKU:
recheck

jpwhitemn (approved):

PR #30, opened 2018-06-15
Double sleeps in edgex deploy script

JPWKU:
This is intented to help debug intermittent
test failure issues where a container service
isn't ready.

### DIFF
--- a/deploy-edgeX.sh
+++ b/deploy-edgeX.sh
@@ -32,41 +32,42 @@ run_service () {
 }
 
 run_service volume
-sleep 10
+sleep 20
 run_service consul
+sleep 20
 
 while ! $(docker-compose exec -T consul nc -z localhost 8500);do echo "not already startup… wait for 5 second reconnect." ;sleep 5; done
 
 run_service config-seed
 run_service mongo
 
-sleep 10
+sleep 20
 
 run_service mongo-seed
 run_service logging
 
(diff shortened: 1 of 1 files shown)


JPWKU:
@tsconn23 @jpwhitemn @cloudxxx8 I just found this https://forums.docker.com/t/solved-docker-compose-exec-error-no-container-found-for-web-1/25828  which is the exact error we are seeing in the failures.  It isn't even making it to the part where the services are (un)available.  I still think we could benefit from the ```depends_on``` directive as a first gate for deploying edgex.

jpwhitemn:
Jeremy - I don't think this is the same issue as I don't think I've seen anyone doing an exec.

    On Friday, June 15, 2018, 3:42:15 PM CDT, Jeremy Phelps <notifications@github.com> wrote:  
 
 
@tsconn23 @jpwhitemn @cloudxxx8 I just found this https://forums.docker.com/t/solved-docker-compose-exec-error-no-container-found-for-web-1/25828 which is the exact error we are seeing in the failures. It isn't even making it to the part where the services are (un)available. I still think we could benefit from the depends_on directive as a first gate for bring the services up.

JPWKU:
@jpwhite Here is where it's getting caught up as far as I can tell https://github.com/edgexfoundry/blackbox-testing/blob/master/deploy-edgeX.sh#L38.
exec's are used in blackbox

PR #35, opened 2018-06-25
add support-logging test script

weichou1229:


weichou1229:
@tsconn23 please help check this PR .

JPWKU:
recheck

andyf1967:
recheck

bin/postman-test/DataDumps/logging/logEntry.js
```diff
@@ -0,0 +1,192 @@
+var dbName = 'logging';
```
tsconn23: Please make this a JSON file. All of the other DataDumps files are JSON files, so this one should be as well. In addition, the other DataDumps directories have a populate.sh script that is used for population. Please use the same methodology in this case.

bin/postman-test/environment/local_edgex.postman_environment.json
```diff
@@ -0,0 +1,42 @@
+{
```
tsconn23: Can we both name this file and set the "name" property so that the default name is appropriate when imported into Postman? "support-logging.json" perhaps?

bin/postman-test/environment/edgex.postman_environment.json
```diff
@@ -0,0 +1,42 @@
+{
```
tsconn23: Can we both name this file and set the "name" property so that the default name is appropriate when imported into Postman? "support-logging-docker.json" perhaps?

tsconn23 (dismissed):

bin/postman-test/DataDumps/logging/logEntry.js
weichou1229: In my opinion, JSON file is not maintainable and hard to check syntax error ,and I also use mongo js lib in support-notification test .

bin/postman-test/environment/local_edgex.postman_environment.json
weichou1229: yes , It's ok .

bin/postman-test/environment/edgex.postman_environment.json
weichou1229: yes , It's ok .

bin/postman-test/DataDumps/logging/logEntry.js
weichou1229: I had add populate.sh and environment file , please check .

tsconn23: @weichou1229 Well then you now have a situation where the mechanism to populate data for this one group of tests is different from the others. It needs to be uniform. If you're going to change it for all of them, do that as a separate PR. @cloudxxx8 

And I just noticed that in the PR for the new export-client tests, you used a JSON file. So please make this consistent.
https://github.com/edgexfoundry/blackbox-testing/pull/39/files#diff-f58cab9c339000f0f7727132fd442c65

bin/postman-test/DataDumps/logging/populate.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
```
tsconn23, line 1: @weichou1229 Thanks for adding this but see my other comments regarding how it's a different methodology from the existing test data population @cloudxxx8

bin/postman-test/DataDumps/logging/logEntry.js
weichou1229: ok , I will change all my PR to JSON file

bin/postman-test/DataDumps/logging/populate.sh
weichou1229: @tsconn23 I had changed dump file to json format , please check .

tsconn23 (approved):
Note additional commit I added at the end during final round of verification. I have all 94 tests passing now via Postman, 0 failures. Merging.

PR #31, opened 2018-06-17
add support-notification test script

weichou1229:
Signed-off-by: Bruce <weichou1229@gmail.com>

PR #42, opened 2018-06-28
Remove mongoSeed

JPWKU:

### DIFF
--- a/bin/env.sh
+++ b/bin/env.sh
@@ -5,7 +5,6 @@ export consul=consul:1.1.0
 export coreMetadata=nexus3.edgexfoundry.org:10004/docker-core-metadata-go
 export coreConfigseed=nexus3.edgexfoundry.org:10004/docker-core-config-seed-go:master
 export mongo=edgexfoundry/docker-edgex-mongo
-export mongoSeed=edgexfoundry/docker-edgex-mongo-seed
 export supportLogging=nexus3.edgexfoundry.org:10004/docker-support-logging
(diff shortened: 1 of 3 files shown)


JPWKU:
Currently testing this out locally.  @feclare was under the impression that seed functionality is in the mongo image now.  I am not sure.

JPWKU:
This is not working locally.

weichou1229:
@JPWKU you can see
https://wiki.edgexfoundry.org/display/FA/Get+EdgeX+Foundry+-+Users
Consul is now incorporated in the config-seed container.  It is no longer needed.
so you can remove it too .

also , this is new docker-compose file for run go version service , it's should update , too.
https://github.com/edgexfoundry/developer-scripts/blob/master/compose-files/docker-compose-california-0.5.2.yml

tsconn23:
@weichou1229 Hi Bruce, actually we split Consul back out into its own container for the 0.6.0 release. Long story, but it improves the ability to test the config seed locally on a developer's machine using Docker before they check in code.

Also, @jpwhitemn has created a 0.6.0 version of the docker-compose file
https://github.com/edgexfoundry/developer-scripts/blob/master/compose-files/docker-compose-california-0.6.0.yml

weichou1229:
@JPWKU please reference my PR https://github.com/edgexfoundry/blackbox-testing/pull/31 ,
I had migrate docker-compose file to 0.6.0 release , you can update these file :
* docker-compose.yml
* deploy-edgeX.sh
* bin/env.sh
* bin/arm64_env.sh  

after update , it can pass the test verify

PR #52, opened 2018-07-16
fix test description and test script not consistence

weichou1229:


weichou1229:
@tsconn23 please check , I modify some test script ,because test description and test script not consistence ,   
`[\"Response time is less than \"+data.responseTime] = responseTime < 200;`  should be `[\"Response time is less than \"+data.responseTime] = responseTime < data.responseTime;`

Issue #74, opened 2018-09-17
Remove Content-Type checking from Put AdminState and OperatingState API in core-command blackbox test


weichou1229:
According to https://github.com/edgexfoundry/edgex-go/blob/master/api/raml/core-command.raml ,
* /device/{id}/opstate/{opState}
* /device/name/{name}/opstate/{opState}
* /device/{id}/adminstate/{adminState}
* /device/name/{name}/adminstate/{adminState}
These four apis no need check response content-type

cloudxxx8:
fixed by https://github.com/edgexfoundry/blackbox-testing/pull/75

PR #75, opened 2018-09-17
Remove Content-Type checking from Put AdminState and OperatingState API in core-command blackbox test

weichou1229:
https://github.com/edgexfoundry/blackbox-testing/issues/74


cloudxxx8 (approved):
LGTM

Issue #76, opened 2018-09-21
Update docker-compose to Version 3


tsconn23:
Both @JPWKU and I strongly recommend updating the blackbox test docker-compose file to version 3 from version 2. The rationale is two-fold:

1.) The California 0.6.1 release docker-compose uses version 3 as will releases going forward
2.) The new ARM builders we are provisioning will be running recent versions of docker CE and docker-compose. We would like to avoid some of the incompatibility issues we've had in the past that required manual action during the release.

cloudxxx8:
@weichou1229 @cherrycl
Get Bruce and Cherry aware

tsconn23:
@weichou1229 @cherrycl @cloudxxx8 There was a PR https://github.com/edgexfoundry/blackbox-testing/pull/77 that was created for this and then closed without review or merge. What was the reason?

weichou1229:
PR #77 still have some error and cherry have vacation in these days. So I will check PR and create new one.

cloudxxx8:
fixed by #80

PR #80, opened 2018-10-02
update docker compose file to v3

weichou1229:
https://github.com/edgexfoundry/blackbox-testing/issues/76


weichou1229:
@JPWKU 
please help check, it seems docker-compose not support v3 file in arm64
```
Version in "./docker-compose.yml" is unsupported. You might be seeing this error because you're using the wrong Compose file version. Either specify a version of "2" (or "2.0") and place your service definitions under the `services` key, or omit the `version` key and place your service definitions at the root of the file to use version 1.
```
https://jenkins.edgexfoundry.org/job/blackbox-verify-cavium-arm64-master/139/console

JPWKU:
I am looking into this, I have a new arm64 image to test out soon.  I'll ping here when I have a job created in the jenkins sandbox.

JPWKU:
@weichou1229 Can you get in touch with the folks at Cavium and find out which docker-compose and docker  versions are running?

weichou1229:
I found a file which you add on ci-management.
https://github.com/edgexfoundry/ci-management/blob/master/jenkins-config/clouds/openstack/Primary/ubuntu18.04-docker-arm64-4c-2g.cfg
I think you can use the latest version. It seems support ubuntu arm64.
Official document:
https://docs.docker.com/install/#supported-platforms
https://docs.docker.com/compose/compose-file/

feclare:
The versions of docker and docker compose:
```
docker-compose --version
docker-compose version 1.8.0, build unknown
```
```
docker version
Client:
Version: 17.03.2-ce
API version: 1.27
Go version: go1.8.3
Git commit: f5ec1e293
Built: Thu Nov 30 14:20:39 2017
OS/Arch: linux/arm64

Server:
Version: dev
API version: 1.36 (minimum version 1.12)
Go version: go1.9.3
```

tsconn23:
recheck

tsconn23 (approved):
blackbox verify passed on new LF ARM infra. Looks good to me! Good job folks!

Issue #83, opened 2018-10-02
support-logging -- eliminate failure test for logs exceeding limit


tsconn23:
This test is obsolete once the following PR is merged into edgex-go:

<img width="635" alt="screen shot 2018-10-02 at 2 24 53 pm" src="https://user-images.githubusercontent.com/1342980/46372312-680a7800-c650-11e8-95d5-2e56204044e9.png">

https://github.com/edgexfoundry/edgex-go/pull/617

This is because a change had been made whereby if the specified limit to the number of log messages (say, 1000) is greater than the configured ReadMaxLimit (say, 100) then the parameter supplied by the user will be changed to agree with ReadMaxLimit. Therefore an error to say there are more logs available than the user specified will never be thrown.

weichou1229:
Rewrite test to check log messages size, which not exceed the configured ReadMaxLimit.

PR #84, opened 2018-10-02
Removal of logging limit exceeded test

tsconn23:
#83 

Local testing results from this PR using services from edgexfoundry/edgex-go#617
screen shot 2018-10-02 at 2 39 45 pm


weichou1229 (approved):

cloudxxx8:
@tsconn23 may we still keep this test and modify the content to test the scenario you mention in the issue?  The test result become 200 OK and confirm we can retrieve 100 records although the users request 101.  That't the test for the new change(edgexfoundry/edgex-go#617).  What do you think?
@weichou1229 could modify the test if you agree

tsconn23:
@cloudxxx8 Sure, that sounds fine. If you're going to make the change, then you can close this PR.

cloudxxx8:
@tsconn23 Thanks, we will do that
@weichou1229 Please help follow up

PR #85, opened 2018-10-08
Rewrite test to check log messages size, which not exceed the configured ReadMaxLimit.

### LINKED ISSUE #83: support-logging -- eliminate failure test for logs exceeding limit

weichou1229:
fix https://github.com/edgexfoundry/blackbox-testing/issues/83

### DIFF
--- a/bin/importLoggingDataDump.sh
+++ b/bin/importLoggingDataDump.sh
@@ -18,7 +18,7 @@ DUMP_FILES=( $LOGGINGDATADUMP )
 
 for index in "${!DUMP_FILES[@]}"
 do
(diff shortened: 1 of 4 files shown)


cloudxxx8 (approved):
LGTM

PR #99, opened 2018-11-09
Fix scripts which cannot detect failed test cases

### LINKED ISSUE #97: Some scripts cannot detect failed test cases
When I done the blackbox-testing for Delhi, I found some script cannot detect failed test cases.
For example,
image
If the result is not 423, it is still passed.  Some errors were passed by this kind of script, so there are some real issues we haven't recorded.

In this test, the following should be indicated as failed but passed
_core-data_
image
_core-metadata_
image
image
_core-command_
image
image

It might be better to review other test scripts one by one.

FelixTing:
Fix #97 
Also update collection format version to 2.1


FelixTing:
@weichou1229 please help review, thanks

FelixTing:
recheck

JPWKU:
recheck

cloudxxx8:
please fix this issue on master only

PR #102, opened 2018-11-12
Fix scripts which cannot detect failed test cases

### LINKED ISSUE #97: Some scripts cannot detect failed test cases

FelixTing:
Fix #97

Also update collection format version to 2.1

### DIFF
--- a/bin/postman-test/collections/core-command.postman_collection.json
+++ b/bin/postman-test/collections/core-command.postman_collection.json
@@ -1,400 +1,983 @@
 {
-  "id": "3b5ccf5c-51dc-8d94-641d-1969bca94d9b",
-  "name": "core-command",
-  "description": null,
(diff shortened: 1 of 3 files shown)


cloudxxx8:
@weichou1229 Could you please help run this PR and confirm the result is the same in your environment?

weichou1229:
Test result is the same in my environment.

cloudxxx8 (approved):
LGTM

PR #103, opened 2018-11-12
Remove sleep when starting the services for black box test v1.0.0

FelixTing:


cloudxxx8:
@FelixTing you should also remove the while loop and echo in the script

FelixTing:
@cloudxxx8 thanks for the reminder

FelixTing:
operated error when squashing commits, I will reopen a PR

Issue #108, opened 2018-11-23
Core-Data Test Changes For Issue #811


tsconn23:
I am creating this issue to track changes that will need to be made to the core-data blackbox tests based on the following issue:
https://github.com/edgexfoundry/edgex-go/issues/811

I have already found one:
`"name": "95 http://localhost:48080/api/v1/event "`

The following check in this test is invalid since IDs are now strings, the content of which for an event should always be based on a UUID (36 chars).

```
"    //Test body value",
"    tests[\"Is Body Conatins Id\"] = responseBody.length === 24;",
```

Rather than enumerating all of the specific changes that need to be made, I hope the forthcoming PR will be indicative.

PR #109, opened 2018-11-26
Core-data test mods, IDs now in UUID format

tsconn23:
#108 

Companion changes for the following edgex-go PR
https://github.com/edgexfoundry/edgex-go/pull/856

I have tested all core-data blackbox tests against the above PR and made changes where appropriate.


cloudxxx8 (approved):
LGTM

Issue #110, opened 2018-11-28
Update Consul version to v1.3.1 in Docker Shell Scripts


tsconn23:
In accordance with the following PR from developer-scripts

https://github.com/edgexfoundry/developer-scripts/issues/85

We'd like to start using Consul v1.3.1 for Edinburgh, so this needs to be included with the blackbox test functionality

PR #111, opened 2018-11-28
Rev'ed Consul image version to v1.3.1

tsconn23:
#110 


JPWKU (approved):
LGTM, we can proxy the x86 through nexus as well if we want. Should be a tad more efficient since there is a high speed link between this nexus and jenkins.

PR #117, opened 2018-12-13
Blackbox test changes for edgex-go PR: #933

jdharms:
Blackbox tests for core-command shouldn't expect Content-Type header to be set for methods with empty response bodies.


brandonforster (approved):
Verified working.

PR #129, opened 2019-01-05
Issue 128 - Update core-metadata device service test result for UUID return length

michaelestrin:
Issue 128 - Update core-metadata device service test result for UUID return length

* Updates length of expected test result to accommodate change from bson.ObjectId.Hex() to uuid.


cloudxxx8 (approved):
LGTM

PR #130, opened 2019-01-07
Predefine the json schema of DeviceService in pre-request script

### LINKED ISSUE #127: core-metadata/deviceservice test 224 failure
The following test is failing due to the URL below being sent to the metadata service

- 224 GET http://localhost:48081/api/v1/deviceservice/{{getDeviceServiceById}}

The key shown in the URL above is not present in the data json file and appears to be populated via javascript. The `{{getDeviceServiceById}}` value relies on a previous test 210 which has an error. I can duplicate the issue by doing the following:

1.) Clean Mongo
2.) Run the core-metadata-importer collection
3.) Under "Collections" on the left, expand the deviceservice folder and double click test 210
4.) In the right hand panel, change the URL to the following: `{{baseUrl}}/api/v1/deviceservice/name/Common room thermostat deviceservice`
5.) Hit the "Send" button
6.) At the bottom, "Test Results" will have the following error
`There was an error in evaluating the test script:  TypeError: Cannot read property '$ref' of undefined`

In the console output, I see a successful response with the device service properties followed by the `TypeError` message. I would try to address this myself, but I'm unsure what the approach should be.

weichou1229:
Fix #127


tsconn23:
The previous error from test 210 related to the `$ref` is gone now but the variable we need to populate in order to perform test 224 is still not getting resolved. See screenshot below from Postman console.

screen shot 2019-01-07 at 5 16 27 pm

weichou1229:
@tsconn23 I added test data at re-request script and set id to {{getDeviceServiceById}}, so it can work now.
And I think prepare test data at re-request script is more manageable and maintainable.

```
var baseUrl = pm.environment.get("baseUrl");
var randName = "addressable-"+ Math.random().toString(36).substring(5);
var addressable = {
  ...
};
var deviceService = {
  ...
};

var request = {
  url: baseUrl+'/api/v1/addressable',
  method: 'POST',
  body: {
    mode: 'raw',
    raw: JSON.stringify(addressable)
  }
};

pm.sendRequest(request, function (err, res) {
    request.url = baseUrl+'/api/v1/deviceservice';
    request.body = JSON.stringify(deviceService);
    pm.sendRequest(request, function (err, res) {
        if (err) {
            console.log(err);
        } else {
            pm.environment.set("getDeviceServiceById", res.stream.toString());
        }
    });
});
```

tsconn23 (approved):

PR #132, opened 2019-01-09
Add "Present test result using Allure framework"

FelixTing:
Updating document to instruct how to run Allure dashboard locally with
black box test


FelixTing:
Commit is not being verified. I will create PR again.

PR #135, opened 2019-01-10
Change expected DeviceProfile ID length

jdharms:
As part of Edgex-Go #878 ( https://github.com/edgexfoundry/edgex-go/pull/999 ), Ids are now UUIDs by default, not BSON ObjectIds.  This change adjusts the expected length of a created DeviceProfile's Id.

Signed-off-by: Daniel Harms <jdharms@gmail.com>


cloudxxx8 (approved):
LGTM

Issue #139, opened 2019-01-21
No container name in docker-compose file


cherrycl:
Container name auto-generated and displayed like "blackbox-testing_metadata_1", if no container name in docker-compose file.

Example:
  ```
  metadata:
    image: ${coreMetadata}
    ports:
    - "48081:48081"
    hostname: edgex-core-metadata
    networks:
      edgex-network:
        aliases:
        - edgex-core-metadata
    volumes:
    - db-data:/data/db
    - log-data:/edgex/logs
    - consul-config:/consul/config
    - consul-data:/consul/data
    depends_on:
    - logging
```

Container status
```
CONTAINER ID        IMAGE                                                                 COMMAND                  CREATED             STATUS              PORTS                                                                                                            NAMES
64cc57523b86        nexus3.edgexfoundry.org:10004/docker-core-metadata-go:0.6.0           "/core-metadata --co…"   12 days ago         Up 2 hours          0.0.0.0:48081->48081/tcp, 48082/tcp                                                                              blackbox-testing_metadata_1
```

PR #140, opened 2019-01-22
Added container_name and removed unused comment line

cherrycl:
Added container_name to all edgex service and removed some of device services.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,7 +11,7 @@ volumes:
 services:
   volume:
     image: ${volume}
-    container_name: edgex-volume
+    container_name: edgex-files
     networks:
     - edgex-network
     volumes:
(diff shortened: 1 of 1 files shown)


cloudxxx8:
fix #139

docker-compose.yml
```diff
@@ -11,6 +11,7 @@ volumes:
 services:
   volume:
     image: ${volume}
+    container_name: edgex-volume
```
tsconn23: Recommend this be named "edgex-files" for consistency with release docker-compose files
https://github.com/edgexfoundry/developer-scripts/blob/master/compose-files/docker-compose-delhi-0.7.1.yml

cherrycl: @tsconn23 I had updated the container_name of volume to "edgex-files". Thanks for your advice.

tsconn23 (approved):

PR #167, opened 2019-02-25
Adding securityservice testing cases.

tingyuz:
Signed-off-by: Tingyu Zeng <tingyu.zeng@rsa.com>

Signed-off-by: Tingyu Zeng <tingyu.zeng@rsa.com>

### DIFF
--- a/enableSecurity.sh
+++ b/enableSecurity.sh
@@ -1,0 +1,2 @@
+#!/bin/bash
+export SECURITY_SERVICE_NEEDED=true
--- a/bin/cleanSecurityserviceTest.sh
+++ b/bin/cleanSecurityserviceTest.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+echo "Info: Clean Securityservice's test data."
+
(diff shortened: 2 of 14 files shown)


enableSecurity.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+export SECURITY_SERVICE_NEEDED=true
```
JPWKU, line 2: the env var is set by the jenkins job, I'm not sure we need it here?

JPWKU (changes requested):

enableSecurity.sh
tingyuz: This script is not being called anywhere currently and it will be used for manual testing. E.g, I need to run this script, then /bin/env.sh, then setupSecurityservice.sh etc to check where might go wrong for degbugging.

JPWKU: Ok that makes sense.

JPWKU (approved):

PR #188, opened 2019-03-29
Remove core-data test #27 issue #187

tsconn23:
#187 

Signed-off-by: Trevor Conn <trevor_conn@dell.com>

PR #190, opened 2019-04-01
Remove duplicate paths for core-data/readings by device and value descriptor

### LINKED ISSUE #187: Remove "Readings by Device and Value Descriptor up to X Results" test
The following test in the core-data/event suite should be removed:
27 - http://localhost:48080/api/v1/event/device/:deviceId/valuedescriptor/{valuedescriptor}/{limit} 

There is a pending PR for the removal of this endpoint https://github.com/edgexfoundry/edgex-go/pull/1209

For additional context as to why this is being removed, please see this issue:
https://github.com/edgexfoundry/edgex-go/issues/1194

Also, the core-data Client function that would allow this endpoint to be called is not part of the [ReadingClient](https://github.com/edgexfoundry/go-mod-core-contracts/blob/ed80a34a1c6e6b174ed14ce3e992c0eb507f9050/clients/coredata/reading.go#L30) interface currently, so it's uncallable except via a direct REST invocation like Postman. See the following issue https://github.com/edgexfoundry/go-mod-core-contracts/issues/45

weichou1229:
Fix #187

### DIFF
--- a/bin/postman-test/collections/core-data.postman_collection.json
+++ b/bin/postman-test/collections/core-data.postman_collection.json
@@ -1,6 +1,6 @@
 {
 	"info": {
-		"_postman_id": "b53ec894-ecb6-44f7-937b-f9602df24bc2",
+		"_postman_id": "88cc25f6-c2ed-4f44-afcb-5e244ddcd2d6",
 		"name": "core-data",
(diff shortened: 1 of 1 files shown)


weichou1229:
Duplicate with https://github.com/edgexfoundry/blackbox-testing/pull/188

PR #194, opened 2019-04-03
Remove _class from test scripts

### LINKED ISSUE #177: Remove _class from the test script
`_class` is the legacy script, and I've confirmed it is useless now. The original test data was copied from MongoDB output and the data was stored by old Java services at that time. The Java OR mapping library generates this additional field in MongoDB to indicate which Java class should be used to deserialize the record.

cherrycl:
fix #177


cloudxxx8 (approved):
LGTM

PR #199, opened 2019-04-10
Remove interval/intervalAction on metadata import/clean file

### LINKED ISSUE #195: core-metadata-importer fails on SchedulerURL
It looks like the core-metadata-importer collection is still trying to import data for the scheduler, which should no longer be the case.
```
"listen": "test",
	"script": {
			"id": "47de7093-d096-4c92-83f4-3f06215ecf31",
			"exec": [
					"var schedulerUrl = pm.environment.get(\"schedulerUrl\");",
```
Screen Shot 2019-04-03 at 12 27 00 PM

Interval and IntervalAction are owned by the scheduler, not metadata

### LINKED ISSUE #196: core-metadata-cleaner still trying to clean schedule data
See intervalAction step below.
Screen Shot 2019-04-03 at 12 36 49 PM

I imagine the collection will go on to clean `interval` and it shouldn't do that either. Interval and IntervalAction are handled within the scheduler service, not metadata

cherrycl:
fix #195 
fix #196


cloudxxx8 (approved):
LGTM

Issue #168, opened 2019-02-26
Create tests for the Go Virtual Device Service


cloudxxx8:
assigned to @FelixTing , please write down the scenarios to the comments of this issue

We will create Go Virtual Device Service for Edinburgh release, and it could be a good target to verify the basic behaviors of Go Device SDK.

This issue will be implemented after the image of Go Virtual Device Service ready

FelixTing:
1. Test whether the virtual device service exists
`http://[host]:49988/api/v1/ping`

2. Test whether the device profiles has been added to metadata
```
http://[host]:48081/api/v1/deviceprofile/name/...
Random-Bool-Generator
Random-Integer-Generator
Random-UnsignedInteger-Generator
Random-Float-Generator
```
3. Test whether the pre-defined devices in metadata exists 
```
http://[host]:48081/api/v1/device/name/...
Random-Bool-Generator01
Random-Integer-Generator01
Random-UnsignedInteger-Generator01
Random-Float-Generator01
```
4. Test the put method of Enable_Randomization and Value
```
Method: PUT
URL: http://[host]:49988/api/v1/device/name/Random-Integer-Generator01/RandomValue_Int8
Header: application/json
Body: {"Enable_Randomization":"false","RandomValue_Int8":"66"}
```
5. Test the get value method
```
http://[host]:49988/api/v1/device/name/Random-Integer-Generator01/RandomValue_Int8
Expected value: 66
```

cloudxxx8:
For the first scenario, you should test whether the Device Service exists, not addressable
http://[host]:48081/api/v1/deviceservice/name/device-virtual

FelixTing:
Oops, I should ping it
http://[host]:49988/api/v1/ping
Thanks @cloudxxx8 
> For the first scenario, you should test whether the Device Service exists, not addressable
> http://[host]:48081/api/v1/deviceservice/name/device-virtual

cloudxxx8:
Reopen this issue.
We will close this one until we have device-virtual-go image and include the test into run.script

cloudxxx8:
fixed #213

PR #213, opened 2019-05-06
Update device-virtual postman_collection

FelixTing:
1. Corrected device name for Random-Boolean-Generator01 in test
description and url
2. Remove the EnableRandomization attribute from put command's parameters,
since EnableRandomization is automatically set to false when issuing a put
value command

refer to #168 


cloudxxx8 (approved):
LGTM

PR #218, opened 2019-05-08
Add response count validation for notification and transmission

### LINKED ISSUE #212: Missing count test for related api on support-notification
Test cases of transmission API which querying with limit always passed, even the response count is incorrect.
An issue has created for incorrect response count on [edgex-go#1312](https://github.com/edgexfoundry/edgex-go/issues/1312).

cherrycl:
fix #212


cloudxxx8 (approved):
LGTM

PR #245, opened 2019-06-27
Change rulesengine start order

### LINKED ISSUE #244: Change rules-engine starting order
We encounter some failed test cases caused by long response time from time to time.  After some investigation, the perfomance might be impacted by rules-engine sevice starting.  Rules-enigne needs more time to start up and consumes ltos of CPU, so it might impact the some test execution if rules-engine is still starting.
A quick solution is move the rules-engine starting earlier, and it could start during downloading other images.

cherrycl:
Change rules-engine started order

Fix #244

### DIFF
--- a/deploy-edgeX.sh
+++ b/deploy-edgeX.sh
@@ -60,15 +60,7 @@ run_service mongo
 
 run_service logging
 
-run_service notifications
-
-run_service metadata
-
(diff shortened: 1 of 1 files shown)


cloudxxx8 (approved):
LGTM

PR #251, opened 2019-07-11
Add tests for new endpoint in metadata

### LINKED ISSUE #250: Add tests for new metadata endpoint restGetCommandsByDeviceId
New metadata endpoint is added: [restGetCommandsByDeviceId](https://github.com/edgexfoundry/edgex-go/blob/03f28e5980c37459c208a296556b0d148bea68db/internal/core/metadata/rest_command.go#L83).
Tests need to be added for that endpoint

difince:
- add tests for restGetCommandsByDeviceId metadata endpoint.
- change slightly some of the tests descriptions
- added fix for the false positive test that tries to get list of commands by name.
Added validation that the return result is empty list, instead of 404.The endpoint
returns list of commands with specified name or empty list if no command with that name exists.

Fix: https://github.com/edgexfoundry/blackbox-testing/issues/250

Signed-off-by: difince <dianaa@vmware.com>

### DIFF
--- a/bin/postman-test/collections/core-metadata.postman_collection.json
+++ b/bin/postman-test/collections/core-metadata.postman_collection.json
@@ -1093,7 +1093,7 @@
 								"{{commandGetRequestId}}"
 							]
 						},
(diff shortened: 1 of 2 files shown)


cherrycl (approved):
LGTM

cloudxxx8 (approved):
LGTM

PR #262, opened 2019-08-30
Fix - Use existing device names when testing commands

### LINKED ISSUE #261: Some Commands Tests use not existing device names
Some command's tests use not existing device names - Random-Boolean-Device. ([example](https://github.com/edgexfoundry/blackbox-testing/blob/bff0f8d98750236b98600e981cd3919a00f35a45/bin/postman-test/collections/core-command.postman_collection.json#L270), [example2](https://github.com/edgexfoundry/blackbox-testing/blob/bff0f8d98750236b98600e981cd3919a00f35a45/bin/postman-test/collections/core-command.postman_collection.json#L426), [example3](https://github.com/edgexfoundry/blackbox-testing/blob/bff0f8d98750236b98600e981cd3919a00f35a45/bin/postman-test/collections/core-command.postman_collection.json#L270) ), or is not set [correctly](https://github.com/edgexfoundry/blackbox-testing/blob/bff0f8d98750236b98600e981cd3919a00f35a45/bin/postman-test/collections/core-command.postman_collection.json#L845-L870) - > incompatibility between setting [DeviceName](https://github.com/edgexfoundry/blackbox-testing/blob/bff0f8d98750236b98600e981cd3919a00f35a45/bin/postman-test/collections/core-command.postman_collection.json#L847) , but expecting {{[getDeviceByName](https://github.com/edgexfoundry/blackbox-testing/blob/bff0f8d98750236b98600e981cd3919a00f35a45/bin/postman-test/collections/core-command.postman_collection.json#L859)}}

-  Replace with device names with existing one. Be sure that the tests do not fails with 404, when not really expecting.

difince:
Fix https://github.com/edgexfoundry/blackbox-testing/issues/261
The tests for endpoints: 
GET/PUT /api/v1/device/name/{devicename}/command/{commandname}
GET/PUT /api/v1/device/{deviceId}/command/{commandId}
used to fail with 404, because not existing device names was used, or 423 because the devices do exists but were in Locked state.
Fixing both,  https://github.com/edgexfoundry/edgex-go/issues/1706 and https://github.com/edgexfoundry/blackbox-testing/issues/261 lead to a side effect - the Endpoints started to fail with status 500. This happens because, now the devices exist (because of the name replacement) and their statuses are all UNLOCKED and the process  actually tries to execute real GET/PUT command against a real Device(Device Service), but such is not alive. The returned error is 500 - "Put http://172.17.0.1:48089/cooling: dial tcp 172.17.0.1:48089: connect: connection refused". Could we mimic somehow a device service, or we should not expect anymore 200 ?

Signed-off-by: difince <dianaa@vmware.com>


cherrycl:
@difince In order to get 200, I used the device "Random-Boolean-Device" which was created for device-virtual service. When running those testcases, the device-virtual should be running.
For testing return code 423, the device name has to use the real device? Or we can use the device which existed on device-virtual?

difince:
@cherrycl

difince:
Thank you, @cherrycl. Now, with the virtual-device service up - the tests 

> For testing return code 423, the device name has to use the real device? Or we can use the device which existed on device-virtual?  

For testing 423 code, we do not care if the device is real, or it just exists in the database. Once https://github.com/edgexfoundry/edgex-go/issues/1706 is fixed, the first thing that will be done when the device is loaded from the DB, will be admin status validation. If the state is Locked-> user request will be immediately rejected with 423 and no command will be sent to the device. That is why in this case we should not care if the device is dummy.

I will close both the [issue](https://github.com/edgexfoundry/blackbox-testing/issues/261) and the PR

cloudxxx8:
@difince this part has been implemented in Device Service SDK, so in core-command, it might only need to pass the status code to the client

> the first thing that will be done when the device is loaded from the DB, will be admin status validation. If the state is Locked-> user request will be immediately rejected with 423 and no command will be sent to the device.

PR #269, opened 2019-09-09
Change the test device for core-command testcase

### LINKED ISSUE #268: Device name of testcase 30_404 should be change to unlocked device
The test device is locked, and it can't be used in this case.

cherrycl:

fix #268


cloudxxx8 (approved):
LGTM

Issue #282, opened 2019-10-21
Blackbox tests failing in Jenkins


michaelestrin:
From @tsconn23:

> Hi all – This is a heads up that I noticed just now that the blackbox tests in Jenkins stopped working on 18-October.
>
>https://jenkins.edgexfoundry.org/view/blackbox-testing/job/blackbox-testing-centos7-blackbox-4c-2g-master/
>
>If you look at the Console Output for one of the failed reports, you’ll see errors like the following
>
>08:00:43 → Test whether the virtual device service exists
>08:00:43   GET http://edgex-device-virtual:49990/api/v1/ping [errored]
>08:00:43      connect ECONNREFUSED 172.17.0.6:49990
>08:00:43 → Test whether the Random-Boolean-Device device profile has been added to metadata
>08:00:43   GET http://edgex-core-metadata:48081/api/v1/deviceprofile/name/Random-Boolean-Device [errored]
>08:00:43      getaddrinfo ENOTFOUND edgex-core-metadata edgex-core-metadata:48081
>08:01:05      getaddrinfo ENOTFOUND edgex-core-data edgex-core-data:48080
>08:01:05 [errored]
> 
>I went through the Jenkins job definition and then through the relevant scripts in the blackbox-testing repo to see if I could find a commit between the 17th and 18th that would be a candidate for causing the failure, but nothing is jumping out at me.
>
>It’s also my understanding that the blackbox tests are using the docker-compose file below and not the one from developer-scripts/releases/nightly-build/compose-files which has seen a lot of modification.
>
>https://github.com/edgexfoundry/blackbox-testing/blob/master/docker-compose.yml

michaelestrin:
PR did not address issue.  Still seeing numerous failures like `connect ECONNREFUSED 172.17.0.5:49990` and `getaddrinfo ENOTFOUND edgex-core-metadata edgex-core-metadata:48081`.

michaelestrin:
When attempting to deploy to my local Windows machine, several containers appear to have issues:

```
Michael@DESKTOP-UM6HQHJ MINGW64 ~/source/dell/projects/blackbox-testing/src/github.com/edgexfoundry/blackbox-testing (issue-282)
$ docker ps -a
CONTAINER ID        IMAGE                                                                  COMMAND                  CREATED             STATUS                     PORTS                                                                                              NAMES
[...]

PR #284, opened 2019-10-21
Blackbox tests failing in Jenkins

michaelestrin:
Corrects Docker image/version differences missed in last PR.  Will
not resolve the issue completely. See notes in issue for details.

https://github.com/edgexfoundry/blackbox-testing/issues/282

### DIFF
--- a/bin/arm64_env.sh
+++ b/bin/arm64_env.sh
@@ -19,10 +19,10 @@ export exportDistro=nexus3.edgexfoundry.org:10004/docker-export-distro-go-arm64:
 export deviceVirtual=nexus3.edgexfoundry.org:10004/docker-device-virtual-go-arm64:1.1.0
 export appService=nexus3.edgexfoundry.org:10004/docker-app-service-configurable-arm64:latest
 
-export vault=nexus3.edgexfoundry.org:10004/docker-edgex-vault-arm64:0.3.0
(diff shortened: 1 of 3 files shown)


bin/arm64_env.sh
```diff
@@ -19,10 +19,10 @@ export exportDistro=nexus3.edgexfoundry.org:10004/docker-export-distro-go-arm64:
 export deviceVirtual=nexus3.edgexfoundry.org:10004/docker-device-virtual-go-arm64:1.1.0
 export appService=nexus3.edgexfoundry.org:10004/docker-app-service-configurable-arm64:latest
 
-export vault=nexus3.edgexfoundry.org:10004/docker-edgex-vault-arm64:0.3.0
-export vaultWorker=nexus3.edgexfoundry.org:10004/docker-edgex-vault-worker-go-arm64:0.3.0
-export kongdb=postgres:9.5
-export kong=kong:1.3.0-ubuntu
```
anonymouse64, line 25: note that this does actually indeed need to be version 1.3, kong 1.2 and below doesn't run on arm64, so to run this at all we need kong 1.3 on arm64. Arguably we should also upgrade to kong 1.3 on amd64 as well

anonymouse64 (changes requested):

michaelestrin:
Closed by accident.  

:(

Issue #287, opened 2019-10-22
core-command device tests missing required parameter


michaelestrin:
Tests 21_200 and 31_200 PUT command to device-virtual RandomValue_Int8 are missing EnableRandomization_Int8 parameter.

michaelestrin:
Found when running tests for https://github.com/edgexfoundry/edgex-go/pull/1972

cloudxxx8:
@michaelestrin did you encounter any issue on this?  According to the design, it would be automatically set to false by default.

michaelestrin:
@cloudxxx8 Tests run but have two failures:

```
C:\Users\Michael\AppData\Roaming\npm\newman.cmd run bin/postman-test/collections/core-command.postman_collection.json -e bin/postman-test/environment/core-command-docker.postman_environment.json --folder device -d bin/postman-test/data/coreCommandData.json --disable-unicode
newman

core-command

Folder device
Sub-folder http://localhost:48082/api/v1/device
  GET http://edgex-core-command:48082/api/v1/device [200 OK, 12.16KB, 401ms]
  Pass  Status code is 200
  Pass  Response time is less than 800
  Pass  Has Content-Type
  Pass  Content-Type is application/json

Sub-folder 17 http://localhost:48082/api/v1/device/:id
  GET http://edgex-core-command:48082/api/v1/device/name/Livingroom automation [200 OK, 746B, 11ms]
  GET http://edgex-core-command:48082/api/v1/device/a776a784-cd7e-427d-831e-5e958e5d07bb [200 OK, 746B, 10ms]
  -
  | 'Validation failed: ', null
  -
  Pass  Status code is 200
  Pass  Response time is less than 1200 ms
  Pass  Has Content-Type
  Pass  Content-Type is application/json
  Pass  Is Valid Response Schema

Sub-folder 20_200 http://localhost:48082/api/v1/device/:id/command/{commandId}
  GET http://edgex-core-command:48082/api/v1/device/name/Random-Boolean-Device [200 OK, 984B, 10ms]
  GET http://edgex-core-command:48082/api/v1/device/7d6e57ad-c2af-4182-87f6-cf79574faf2a/command/d30dab15-6c9a-4c65-a708-6513b480f8c2 [200 OK, 279B, 19ms]
  -
  | 'Validation failed: ', null
  -
  Pass  Status code is 200
  Pass  Response time is less than 800ms
  Pass  Has Content-Type
  Pass  Content-Type is application/json
[...]

PR #288, opened 2019-10-22
core-command device tests missing required parameter

michaelestrin:
Adds missing EnableRandomization_Int8 parameter for tests 21_200 and
31_200.

https://github.com/edgexfoundry/blackbox-testing/issues/287


cloudxxx8:
fixed in device-virtual-go

PR #289, opened 2019-10-22
feat(app-service-configurable) : Update tests for removal of docker profiles

lenny-goodell:
app-service-configurable now uses env var overrides rather than separate docker profiles.

Update compose file to use env var overrides, rather than docker profiles.
Update postman collection to refer to non-docker profile.


cherrycl (approved):
LGTM

cloudxxx8 (approved):

PR #297, opened 2019-10-23
Update to leverage security environment variable

### LINKED ISSUE #294: Parameterize EDGEX_SECURITY_SECRET_STORE in docker-compose.yml
Parameterize the value for EDGEX_SECURITY_SECRET_STORE [here](https://github.com/edgexfoundry/blackbox-testing/blob/master/docker-compose.yml#L22).

Update [deployment](https://github.com/edgexfoundry/blackbox-testing/blob/master/deploy-edgeX.sh) to set the parameterized value based on the value of $SECURITY_SERVICE_NEEDED.

AnthonyMBonafide:
Fix #294

Update deploy script and Docker Compose file to leverage the
'EDGEX_SECURITY_SECRET_STORE' environment variable to disable security.
Also, update the deploy script to match the rest of EdgeX by enabling
security by default unless the 'EDGEX_SECURITY_SECRET_STORE' environment
variable is explicitly set to 'false'.

Signed-off-by: Anthony M. Bonafide <AnthonyMBonafide@gmail.com>

### DIFF
--- a/deploy-edgeX.sh
+++ b/deploy-edgeX.sh
@@ -30,7 +30,14 @@ run_service () {
 	docker-compose up -d $1
 }
 
+if [ "$SECURITY_SERVICE_NEEDED" = "true" ]; then
+	export SECURITY_IS_ON="true"
+else
+	export SECURITY_IS_ON="false"
+fi
+
 run_service volume
+
 run_service consul
 
 run_service config-seed
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,7 +19,7 @@
 version: '3.4'
 # all common shared environment variables defined here:
 x-common-env-variables: &common-variables
-  EDGEX_SECURITY_SECRET_STORE: "false"
+  EDGEX_SECURITY_SECRET_STORE: ${SECURITY_IS_ON}
 
 volumes:
   db-data:


deploy-edgeX.sh
```diff
@@ -31,13 +31,14 @@ run_service () {
...
 run_service consul
 
 run_service config-seed
 
-
-
-if [ "$SECURITY_SERVICE_NEEDED" = "true" ]; then
+# Security is enabled by default unless specified. This matches the rest of
+# the EdgeX system.
+if [ "$EDGEX_SECURITY_SECRET_STORE" != "false" ]; then
```
AnthonyMBonafide: I updated this to only use the `EDGEX_SECURITY_SECRET_STORE` environment variable. This makes it so that we only have to specify the one environment variable rather than two. Also, this makes the setup consistent with the rest of EdgeX.

AnthonyMBonafide:
With these changes we will need to update the Jenkins jobs to include the appropriate value for the environment variable `EDGEX_SECURITY_SECRET_STORE`

michaelestrin (approved):
LGTM

deploy-edgeX.sh
AnthonyMBonafide: Ignore this comment. This has been reverted.

PR #315, opened 2019-11-05
Update banner and version

### LINKED ISSUE #313: The banner words should be updated to EdgeX Foundry
Fuse is the old project code, and it should be updated to match the document.
https://github.com/edgexfoundry/blackbox-testing/blob/master/bin/banner.sh

Document has been updated
https://github.com/edgexfoundry/blackbox-testing/blob/master/docs/How-to-run-blackbox-testing.rst

jinlinGuan:
Fix https://github.com/edgexfoundry/blackbox-testing/issues/313


cloudxxx8 (approved):
LGTM

PR #316, opened 2019-11-05
Removed references to Mongo in prerequisites.

### LINKED ISSUE #312: Rmove Mongo in documentation
We modified the test data import mechanism to avoid database restriction, and we are going to use Redis as default database in the next release.
However, the document still states Mongo as a required service in List of Test Prerequisites section. https://github.com/edgexfoundry/blackbox-testing/blob/master/docs/How-to-run-blackbox-testing.rst

brandonforster:
Fixes #312

Note that the are other references to Mongo in the document. I am too unfamiliar with them to know exactly what it should be updated to.

Signed-off-by: Brandon Forster <brandonforster@gmail.com>

### DIFF
--- a/docs/How-to-run-blackbox-testing.rst
+++ b/docs/How-to-run-blackbox-testing.rst
@@ -353,7 +353,7 @@ Core-Command
 
 To run core-command tests you must have the following services running before any beginning any tests:
 
-1. mongo
+1. database
 2. core-data
 3. core-metadata
 4. core-command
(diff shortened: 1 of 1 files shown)


brandonforster (commented):
@cloudxxx8 Does the reference to Mongo on line 84 also need to be removed?

https://github.com/edgexfoundry/blackbox-testing/blame/master/docs/How-to-run-blackbox-testing.rst#L84

cloudxxx8 (approved):
LGTM

PR #324, opened 2019-11-06
bug(rules engine): Must have App Service Configurable rules profile running (fuji branch)

lenny-goodell:
App Service Configurable running rules-engine profile must be running for the Rules Engines to run without connection errors.

Closes edgexfoundry/edgex-go#2056 (for fuji branch)


lranjbar (approved):
LGTM we need these changes in fuji branch as well

PR #328, opened 2019-11-07
update test case for fuji release

tingyuz:
update unit testing case for security service due to the changes on FUJI release.

Signed-off-by: Tingyu Zeng <Tingyu.Zeng@rsa.com>


lranjbar (approved):
We have these changes in master and need them in fuji as well

Issue #334, opened 2019-11-13
Blackbox test arm64-fuji failure caused by the failed support-notification ping test


jinlinGuan:
<img width="984" alt="Screen Shot 2019-11-13 at 11 23 21 AM" src="https://user-images.githubusercontent.com/55377716/68730626-36f8c700-0608-11ea-98ec-6c8189454782.png">

judehung:
The root cause is due to following error thrown from postman newman container when any assertion errors occurs:

```
/usr/lib/node_modules/newman/node_modules/cli-table3/src/cell.js:60
let fixedWidth = tableOptions.colWidths[this.x];
^
TypeError: Cannot read property '0' of undefined
at Cell.mergeTableOptions (/usr/lib/node_modules/newman/node_modules/cli-table3/src/cell.js:60:44)
at Table.<anonymous> (/usr/lib/node_modules/newman/node_modules/cli-table3/src/table.js:27:14)
at Array.forEach (<anonymous>)
at Table.<anonymous> (/usr/lib/node_modules/newman/node_modules/cli-table3/src/table.js:26:11)
at Array.forEach (<anonymous>)
at Table.toString (/usr/lib/node_modules/newman/node_modules/cli-table3/src/table.js:25:11)
at EventEmitter.<anonymous> (/usr/lib/node_modules/newman/lib/reporters/cli/index.js:90:22)
at EventEmitter.emit (/usr/lib/node_modules/newman/node_modules/eventemitter3/index.js:203:33)
at /usr/lib/node_modules/newman/lib/run/index.js:309:33
at wrapper (/usr/lib/node_modules/newman/node_modules/async/internal/once.js:12:16)
at iteratorCallback (/usr/lib/node_modules/newman/node_modules/async/eachOf.js:56:13)
at /usr/lib/node_modules/newman/node_modules/async/internal/onlyOnce.js:12:16
at wrapper (/usr/lib/node_modules/newman/node_modules/async/dist/async.js:270:20)
at next (/usr/lib/node_modules/newman/node_modules/async/dist/async.js:4560:24)
at /usr/lib/node_modules/newman/node_modules/async/dist/async.js:323:20
at /usr/lib/node_modules/newman/lib/run/export-file.js:41:13
```

Similar error had been reported against newman, but the newman team eventually doesn’t fix this issue: https://github.com/postmanlabs/newman/issues/1904

[...]

PR #335, opened 2019-11-13
Add Echo to support-notification and system-management test scripts

jinlinGuan:
Fix issue 334:https://github.com/edgexfoundry/blackbox-testing/issues/334


jinlinGuan:
Close this PR because the blackbox-testing is not the lastest version.

Issue #295, opened 2019-10-23
Use the appropriate developer-scripts docker-compose file for blackbox tests


michaelestrin:
Created this issue per our conversation earlier today.

Use the appropriate docker-compose file from https://github.com/edgexfoundry/developer-scripts/blob/master/releases/nightly-build/compose-files to start EdgeX.  This will provide a single source of truth for the docker-compose files.

This will require arm64 versions of the docker-compose files be created in developer-scripts.  A separate issue has been created for that work.  https://github.com/edgexfoundry/developer-scripts/issues/154

Once the change is made to use the developer-scripts docker-compose file to bring up EdgeX for blackbox testing, we also discussed the need to modify the deployment to do a `docker-compose up` for the entire system instead of starting individual services one at a time.

Currently, the developer-scripts docker-compose with security has a race condition.  There is an outstanding PR (https://github.com/edgexfoundry/developer-scripts/pull/147) that needs to be validated and merged prior to you working this issue.

This is targeted for Geneva.

michaelestrin:
Suggestion received from @anonymouse64 -- when you make these changes, you may also want to stop depending upon the $SECURITY_SERVICE_NEEDED being set conditionally within the Jenkins build. Instead use the $JENKINS_URL value to see what job is currently running and based on that decide what to do in the script.

> For example, take a look at what we do for the snap job: https://github.com/edgexfoundry/ci-management/blob/master/shell/edgexfoundry-snapcraft.sh#L12-L29.  This is a weird example because that script is in ci-management, but you could have something similar in blackbox-tests repo that checks if $JENKINS_URL matches blackbox-tests-secty-* and then set the env var, etc.

PR #350, opened 2019-12-13
Get docker-compose from developer-script repo

jinlinGuan:
issue #295 

variable RELEASE to select specific release, the default is nightly-build
variable FOR_REDIS=ture to select redis-no-secty.yml
variable SECURITY_SERVICE_NEEDED=true to select security compose file

It successfully run on my local(macOS Catalina), but gets errors on jenkins sandbox.
Running core data test:
image


cloudxxx8:
This PR is working in process.  We can run on local machine, but encounter issue on Sandbox.
We need more time to resolve it.

cloudxxx8:
We have resolved the permission issue, and it could pass on sandbox now

cloudxxx8:
@jinlinGuan please rebase and resolve the conflict

bin/arm64_env.sh
```diff
@@ -1,27 +1,4 @@
...
-export coreMetadata=nexus3.edgexfoundry.org:10004/docker-core-metadata-go-arm64:master
-export coreCommand=nexus3.edgexfoundry.org:10004/docker-core-command-go-arm64:master
-
-export supportLogging=nexus3.edgexfoundry.org:10004/docker-support-logging-go-arm64:master
-export supportNotifications=nexus3.edgexfoundry.org:10004/docker-support-notifications-go-arm64:master
-export supportScheduler=nexus3.edgexfoundry.org:10004/docker-support-scheduler-go-arm64:master
-export supportRulesengine=nexus3.edgexfoundry.org:10004/docker-support-rulesengine-arm64:master
-
-export deviceVirtual=nexus3.edgexfoundry.org:10004/docker-device-virtual-go-arm64:master
 export appService=nexus3.edgexfoundry.org:10004/docker-app-service-configurable-arm64:latest
```
lranjbar: Why isn't this consistent with the rest of the services? This should be pulled from the compose file as well, correct?

bin/env.sh
```diff
@@ -1,28 +1,5 @@
...
-export systemManagement=nexus3.edgexfoundry.org:10004/docker-sys-mgmt-agent-go:master
-
-export vault=vault:1.0.3
-export vaultWorker=nexus3.edgexfoundry.org:10004/docker-edgex-security-secretstore-setup-go:master
-export kongdb=postgres:9.6
-export kong=kong:1.3.0-centos
-export edgexProxy=nexus3.edgexfoundry.org:10004/docker-edgex-security-proxy-setup-go:master
-export securitySecretsSetup=nexus3.edgexfoundry.org:10004/docker-edgex-secrets-setup-go:master
-
 export postman=postman/newman
```
lranjbar: This version of newman should be pinned to same specific version identifed in the ARM build.

bin/env.sh
```diff
@@ -1,28 +1,5 @@
...
-export coreMetadata=nexus3.edgexfoundry.org:10004/docker-core-metadata-go:master
-export coreCommand=nexus3.edgexfoundry.org:10004/docker-core-command-go:master
-
-export supportLogging=nexus3.edgexfoundry.org:10004/docker-support-logging-go:master
-export supportNotifications=nexus3.edgexfoundry.org:10004/docker-support-notifications-go:master
-export supportScheduler=nexus3.edgexfoundry.org:10004/docker-support-scheduler-go:master
-export supportRulesengine=nexus3.edgexfoundry.org:10004/docker-support-rulesengine:master
-
-export deviceVirtual=nexus3.edgexfoundry.org:10004/docker-device-virtual-go:master
 export appService=nexus3.edgexfoundry.org:10004/docker-app-service-configurable:latest
```
lranjbar: Also should be pulled from the compose file.

lranjbar (changes requested):

bnevis-i:
@jinlinGuan I am diffing developer-scripts and blackbox-test and noticed that blackbox has /etc/newman mounted to several places.  Can you plain this more? What is this volume mount used for?

cloudxxx8:
@bnevis-i It only exists in volume and mongo as I see.  There are some historical reasons, but they are useless right now.  This PR will eliminate those /etc/newman folder from other containers.

bin/arm64_env.sh
cloudxxx8: @lranjbar this is a special case.  There are two application services running for blackbox test.  One is **app-service-rules** and another is **app-service-configurable**.  They are using the same docker image with different configuration because it is **configurable** application service.  The app-service-configurable here is provided for test only, so it doesn't exist in developer-script.
https://github.com/edgexfoundry/developer-scripts/blob/master/releases/nightly-build/compose-files/docker-compose-nexus.yml

You can see the configuration profile here is `blackbox-tests`

```
  app-service-configurable:
    image: ${appService}
    ports:
      - "48095:48095"
    entrypoint: ["/app-service-configurable"]
    command: ["--registry","--confdir=/res", "--profile=blackbox-tests", "--skipVersionCheck=true"]
    container_name: edgex-app-service-configurable
    hostname: app-service-configurable
    environment:
      edgex_registry: consul://edgex-core-consul:8500
      edgex_service: http://app-service-configurable:48095
      edgex_profile: blackbox-tests
      Service_Host: app-service-configurable
      Clients_CoreData_Host: edgex-core-data
      Clients_Logging_Host: edgex-support-logging
      Logging_EnableRemote: "true"
      MessageBus_SubscribeHost_Host: edgex-core-data
      Database_Host: edgex-mongo
      Database_Username: appservice
      Database_Password: password
    networks:
      edgex-network:
        aliases:
          - edgex-app-service-configurable
```

Thus, we need to keep this part in this repository.

bin/env.sh
cloudxxx8: The same reason as mentioned.

bin/env.sh
cloudxxx8: Thanks for the reminder.  @jinlinGuan will update it to use the same version tag as the ARM one.

bin/arm64_env.sh
lranjbar: Then it needs to be updated to use the master tag. There is a discrepancy that shouldn't be there.

sync.sh
```diff
@@ -0,0 +1,24 @@
...
+     export COMPOSE_FILE="releases/${USE_RELEASE}/compose-files/docker-compose-nexus${USE_REDIS}${USE_NO_SECURITY}${USE_ARM64}.yml"
+else
+     export COMPOSE_FILE="releases/${USE_RELEASE}/compose-files/docker-compose${USE_REDIS}-${USE_RELEASE}${USE_NO_SECURITY}${USE_ARM64}.yml"
+fi
+
+echo $COMPOSE_FILE
+git remote add dev https://github.com/edgexfoundry/developer-scripts.git
+git checkout -b devBranch
+git fetch dev
+git checkout dev/master -- $COMPOSE_FILE
```
bnevis-i: Wouldn't it be more straightforward to just 
```
wget "https://raw.githubusercontent.com/edgexfoundry/developer-scripts/master/releases/nightly-build/compose-files/docker-compose-nexus.yml"
```

with a few variable substitutions thrown in?

bnevis-i:
This PR doesn't seem to address the requirement to use kong-centos and kong-ubuntu vs kong-alpine in the CI environment.

bin/arm64_env.sh
jinlinGuan: fixed

bin/env.sh
jinlinGuan: fixed

sync.sh
bnevis-i: Suggested replacement:

```
echo $COMPOSE_FILE
wget -q -O docker-compose.yml "https://raw.githubusercontent.com/edgexfoundry/developer-scripts/master/${COMPOSE_FILE}"

if [ `uname -m` = "aarch64" ]; then
    sed -i 's/kong:1.3.0$/kong:1.3.0-ubuntu/' docker-compose.yml
else
    sed -i 's/kong:1.3.0$/kong:1.3.0-centos/' docker-compose.yml
fi
```

bnevis-i (changes requested):

bnevis-i:
`sync.sh` needs to be `chmod +x`

deploy-edgeX.sh
```diff
@@ -25,9 +25,11 @@
 #    . $(dirname "$0")/bin/env.sh
 #fi
 
+source ./sync.sh
```
bnevis-i: Suggest just execute
```
./sync.sh
```

and make sync.sh executable.

deploy-edgeX.sh
```diff
@@ -25,9 +25,11 @@
...
+
 run_service () {
 	echo "\033[0;32mStarting.. $1\033[0m"
-	docker-compose up -d $1
+        docker-compose -f $COMPOSE_FILE up -d $1
```
bnevis-i, line 33: Unnecessary change if we just dynamically download `docker-compose.yml` and modify it in place.

Note: be sure to add `docker-compose.yml` to .gitignore.

sync.sh
jinlinGuan: Why isn't the kong image changed in developer-script instead? This PR [edgexfoundry/developer-scripts#187](https://github.com/edgexfoundry/developer-scripts/pull/187/files) will add docker-compose files which contain kong:1.3.0-ubuntu for arm64.

bnevis-i: > Why isn't the kong image changed in developer-script instead? This PR [edgexfoundry/developer-scripts#187](https://github.com/edgexfoundry/developer-scripts/pull/187/files) will add docker-compose files which contain kong:1.3.0-ubuntu for arm64.

Good point.

How about this then:
```
echo $COMPOSE_FILE
wget -q -O docker-compose.yml "https://raw.githubusercontent.com/edgexfoundry/developer-scripts/master/${COMPOSE_FILE}"

# Use Centos base image instead of Alpine base image for Kong for x86_64 CI
# due to compatibility issues with Alpine image in CI
if [ `uname -m` = "x86_64" ]; then
    sed -i 's/kong:1.3.0$/kong:1.3.0-centos/' docker-compose.yml
fi
```

deploy-edgeX.sh
jinlinGuan: The original `docker-compose.yml` is still needed, because it has the services for blackbox-testing only.  Thus, it is necessary to use `$COMPOSE_FILE` to specify the compose file from developer-script.

bnevis-i: I see now.  Perhaps call them `docker-compose-services-stack.yml` and `docker-compose-test-tools.yml` to distinguish the two?  (This would also eliminate the annoying merge conflicts in the meantime.)

sync.sh
jinlinGuan: Changed to:
```
wget -q "https://raw.githubusercontent.com/edgexfoundry/developer-scripts/master/releases/${USE_RELEASE}/compose-files/${COMPOSE_FILE}"

# Use Centos base image instead of Alpine base image for Kong for x86_64 CI
# due to compatibility issues with Alpine image in CI
# sed command of MacOS is in different syntax 
[ "$(uname -m)" = "x86_64" ] && [ "$(uname)" = "Darwin" ] && sed -i '' 's/kong:1.3.0$/kong:1.3.0-centos/' ${COMPOSE_FILE} || sed -i 's/kong:1.3.0$/kong:1.3.0-centos/' ${COMPOSE_FILE}

```

deploy-edgeX.sh
jinlinGuan: fixed

deploy-edgeX.sh
jinlinGuan: Good idea!

The original `docker-compose.yml` is changed to `docker-compose-test-tools.yml`, however, the compose file from developer-scripts still uses the name as its name in developer-scripts to make sure using the correct compose file as expected.

bin/appServiceConfigurableTest.sh
```diff
@@ -19,7 +19,7 @@ echo "Info: Initiating AppServiceConfigurable Test."
 
 echo "[info] ---------- use docker-compose run newman ----------"
 
-docker-compose run --rm postman run ${COLLECTION_PATH} \
+docker-compose -f ../docker-compose-test-tools.yml docker-compose-test-tools.yml run --rm postman run ${COLLECTION_PATH} \
```
bnevis-i: Not sure what is going on here with the duplication of `docker-compose-test-tools.yml`.  What is this doing?

sync.sh
```diff
@@ -0,0 +1,31 @@
...
+else
+     COMPOSE_FILE="docker-compose${USE_REDIS}-${USE_RELEASE}${USE_NO_SECURITY}${USE_ARM64}.yml"
+fi
+
+wget -q "https://raw.githubusercontent.com/edgexfoundry/developer-scripts/master/releases/${USE_RELEASE}/compose-files/${COMPOSE_FILE}"
+
+# Use Centos base image instead of Alpine base image for Kong for x86_64 CI
+# due to compatibility issues with Alpine image in CI
+# sed command of MacOS is in different syntax 
+[ "$(uname -m)" = "x86_64" ] && [ "$(uname)" = "Darwin" ] && sed -i '' 's/kong:1.3.0$/kong:1.3.0-centos/' ${COMPOSE_FILE} || sed -i 's/kong:1.3.0$/kong:1.3.0-centos/' ${COMPOSE_FILE}
```
bnevis-i: I don't think this line does what you want.  if any of the condtions fails (x86_64, Darwin, or sed) it will end up running the Linux sed command instead of the BSD one.   At a minimum, I would make the x86_64 check into an if/then.

bin/appServiceConfigurableTest.sh
jinlinGuan: It's a typo. Thank you for catching it.

Fixed.

sync.sh
jinlinGuan: Changed to:
```
# Use Centos base image instead of Alpine base image for Kong for x86_64 CI
# due to compatibility issues with Alpine image in CI
# sed command of MacOS is in different syntax 
if [ "$(uname -m)" = "x86_64" ] && [ "$(uname)" = "Darwin" ]; then
     sed -i '' 's/kong:1.3.0$/kong:1.3.0-centos/' ${COMPOSE_FILE}
elif [ "$(uname -m)" = "x86_64" ]; then
     sed -i 's/kong:1.3.0$/kong:1.3.0-centos/' ${COMPOSE_FILE}
fi
```

cloudxxx8:
@lranjbar please help review again when you are back.  thanks

bnevis-i:
Blocking issue: docker-edgex-consul was not rebuilt after https://github.com/edgexfoundry/docker-edgex-consul/pull/16 was merged:
 https://nexus3.edgexfoundry.org/#browse/search/docker=format%3Ddocker%20AND%20attributes.docker.imageName%3Ddocker-edgex-consul:912d0fe7b8192392cfe750c17f11e329:a1f3edd8e1b06cfadf6408e1d7903880 (last updated: Thu Dec 12 2019 09:20:57 GMT-0800 (Pacific Standard Time)

sync.sh
```diff
@@ -0,0 +1,33 @@
...
+
+# nightly or other release
+USE_RELEASE=${RELEASE:-nightly-build}
+if [ "$USE_RELEASE" = "nightly-build" ]; then
+     COMPOSE_FILE="docker-compose-nexus${USE_REDIS}${USE_NO_SECURITY}${USE_ARM64}.yml"
+else
+     COMPOSE_FILE="docker-compose${USE_REDIS}-${USE_RELEASE}${USE_NO_SECURITY}${USE_ARM64}.yml"
+fi
+
+wget -q "https://raw.githubusercontent.com/edgexfoundry/developer-scripts/master/releases/${USE_RELEASE}/compose-files/${COMPOSE_FILE}"
```
bnevis-i: On my system, without the `-O "${COMPOSE_FILE}"` option, when I re-run `deploy-edgeX.sh`  I get multiple dot versions of the compose file download to my system, which creates problems later.

jinlinGuan: added

bnevis-i:
Holding off on further validation until docker-edgex-consul Jenkins pipeline is fixed.  Looks nearly ready to merge.

bnevis-i (approved):
docker-edgex-consul container was published on Dec 27 and resolves all remaining issues.

IMO this is ready to merge.

LGTM

bnevis-i:
@lranjbar Please re-review.

lranjbar (approved):
LGTM :+1: Sorry for the delay because of the holidays.

cloudxxx8:
thanks for all you review!

PR #351, opened 2019-12-17
Fix transmission test cases

jinlinGuan:
addressing the case that the records are less than the limit


jinlinGuan:
close this PR and open another one

PR #353, opened 2019-12-17
Fix operator in the transmission test cases

### LINKED ISSUE #352: Transmission testcases failed when the records are less than the limit
When obtaining limited transmissions by status, if the actual number of records is less than the limit, the test cases will fail. Because in the test cases, it is expected to get the response count exactly the same as the limit.

jinlinGuan:
Fix https://github.com/edgexfoundry/blackbox-testing/issues/352

To address the case that the records are less than the limit, the test is changed to be below the limit.


cloudxxx8 (approved):
LGTM

PR #358, opened 2019-12-19
Fix transmission test cases operator for "equal"

### LINKED ISSUE #357: Transmission testcases failed when the records are equal to the limit
#353 "to.be.below" only addressed the less records, but not the equal records.

jinlinGuan:
Fix https://github.com/edgexfoundry/blackbox-testing/issues/357

change the operator to be at most the limit


cloudxxx8 (approved):
LGTM

PR #361, opened 2019-12-19
Add SElinux sharing labels to all shared volume mounts

bnevis-i:
Add SElinux sharing labels to all *shared* volumes in
order to prevent future issues with blackbox-testing on
SElinux-enabled machines. This issue caused blackbox
testing to fail with security for Genvea release and
s partially fixed. This commit applies this change
universally.  Although this change is mainly
for host-mounted volumes, issues have been observed
for regular docker volumes as well, particularly
for dynamically created directories.

Additionally, set read-only flag on several volumes
that don't require writable shared access.

See related documentation and issues:

https://github.com/docker/docker.github.io/blob/master/storage/bind-mounts.md#configure-the-selinux-label
https://bugzilla.redhat.com/show_bug.cgi?id=1512152

Signed-off-by: Bryon Nevis <bryon.nevis@intel.com>

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -34,9 +34,10 @@ volumes:
   vault-init:
   vault-config:
   vault-file:
-  vault-logs:
-  secrets-setup-cache:
+  vault-logs:  
   newman:
+  # non-shared volumes
+  secrets-setup-cache:
 
 services:
   volume:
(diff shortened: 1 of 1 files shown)


cloudxxx8 (approved):
LGTM
however, after #350 merged, blackbox test will use the docker-compose file in developer-script.
Do we also need to add those sharing labels into the docker-compose file in developer-script?

bnevis-i:
> however, after #350 merged, blackbox test will use the docker-compose file in developer-script.
> Do we also need to add those sharing labels into the docker-compose file in developer-script?

Yes, I have this PR https://github.com/edgexfoundry/developer-scripts/pull/197 filed to backport these changes into developer scripts so that when #350 goes in, the deltas will be minimal.

This other PR (https://github.com/edgexfoundry/docker-edgex-consul/pull/16) is a blocker for security-enabled stack as the services won't come up with the dependency checks in place with the centos or ubuntu based kong images without them.

bnevis-i:
@cloudxxx8 I don't have committer rights in blackbox-testing. Would you do me the favor of merging?

cloudxxx8:
@bnevis-i no problem.  I also triggered the Jenkins job to verify this PR

PR #364, opened 2019-12-20
Enable vault-worker to R/W Vault secrets for all microservices.

### LINKED ISSUE #363: Enable per-service Vault tokens in blackbox-testing
Docker-compose changes are needed to enable https://github.com/edgexfoundry/edgex-go/pull/2253.  Specifically, need to make sure the secrets base directory is writable so that edgex-vault-worker can write all the other services' vault secrets.

bnevis-i:
security-secretstore-setup in conjuction with security-file-token-provider
needs to be able to generate per-service Vault tokens.  Update
the volume mount so that it has R/W access to the basedir of the
secret volumes.

Fixes #363 

Companion PR to https://github.com/edgexfoundry/edgex-go/pull/2253

Signed-off-by: Bryon Nevis <bryon.nevis@intel.com>

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -149,8 +149,7 @@ services:
     volumes:
       - vault-config:/vault/config:z
       - consul-scripts:/consul/scripts:ro,z
-      - /tmp/edgex/secrets/ca:/tmp/edgex/secrets/ca:ro,z
-      - /tmp/edgex/secrets/edgex-kong:/tmp/edgex/secrets/edgex-kong:ro,z
+      - /tmp/edgex/secrets:/tmp/edgex/secrets:z
     depends_on:
       - volume
       - consul


docker-compose.yml
```diff
@@ -149,8 +149,7 @@ services:
     volumes:
       - vault-config:/vault/config:z
       - consul-scripts:/consul/scripts:ro,z
-      - /tmp/edgex/secrets/ca:/tmp/edgex/secrets/ca:ro,z
-      - /tmp/edgex/secrets/edgex-kong:/tmp/edgex/secrets/edgex-kong:ro,z
+      - /tmp/edgex/secrets:/tmp/edgex/secrets:z
```
cloudxxx8, line 152: Should it be `ro, z` for R/W?

bnevis-i: Default is read/write.  "z" flag for shared access.  So just "z" is shared read/write.

bnevis-i: Reason for the change.  Before, security-secrets-setup wrote PKI to /tmp/edgex/secrets. (security-secretstore-setup was just a consumer) In this next release, security-secretstore-setup will also write per-service Vault tokens under here as well (both a producer and consumer).

cloudxxx8 (approved):
LGTM

Issue #332, opened 2019-11-12
Rules engine testing never finished and crashed on blackbox testing ARM platform for Fuji release


tingyuz:
Here is the build job.
https://jenkins.edgexfoundry.org/view/blackbox-testing/job/blackbox-testing-security-ubuntu18.04-docker-arm64-4c-2g-fuji/

The rules engine test case was hanging and never finished. as a result the blackbox testing suite was crashed.

judehung:
From the console output of those recent build failures(build#22 and build#24), both builds failed at the point when the response time of support-notifications ping test exceeds 800ms.

> support-notifications
11:19:09 
11:19:09 ❏ ping
11:19:09 ↳ Test service providing an indication that the service is available.
11:19:10   GET http://edgex-support-notifications:48060/api/v1/ping [200 OK, 105B, 827ms]
11:19:11   ✓  Status code is 200
11:19:11   1. Response time is less than 800ms
11:19:11   ✓  Content-Type is present
11:19:11   ✓  Content-Type is text/plain
11:19:11   ✓  Response body is correct
11:19:12 
11:19:12   #  failure         detail                                                                              
11:19:12                                                                                                          
11:19:12  1.  AssertionError  Response time is less than 800ms                                                    
11:19:12                      expected 827 to be below 800                                                        
11:19:12                      at assertion:1 in test-script                                                       
11:19:12                      inside "ping / Test service providing an indication that the service is available." 
11:19:13 Build step 'Execute shell' marked build as failure

For the successful build#23, the Jenkins job could be executed til the end when the response time of support-notification ping test is below 800ms.

It's strange that the entire Jenkins build was marked as failure for both build#22 and build#24 and skip the remaining testcases execution.

I'm going to setup an arm64 system to see what really happens when the response time of support-notifications ping test fails the assertion.

cloudxxx8:
@lenny-intel could you or any application working group members help on this issue?
after we merge #341 , it sometimes works but still sometimes failed

cherrycl:
Error log in rulesengine

```
14:11:56 #
14:11:56 # A fatal error has been detected by the Java Runtime Environment:
14:11:56 #
14:11:56 #  Internal Error (assembler_aarch64.hpp:1156), pid=6, tid=0x0000ffff7dc26a78
14:11:56 #  guarantee(Rs != Rn && Rs != Rt) failed: unpredictable instruction
14:11:56 #
[...]

PR #365, opened 2019-12-23
Disable rules engine testcases on ARM64 in Fuji

jinlinGuan:
Fix https://github.com/edgexfoundry/blackbox-testing/issues/332

### DIFF
--- a/deploy-edgeX.sh
+++ b/deploy-edgeX.sh
@@ -77,7 +77,7 @@ run_service export-distro
 
 run_service app-service-rules
 
-run_service rulesengine
+[ "$(uname -m)" = "x86_64" ] && run_service rulesengine
 
 run_service notifications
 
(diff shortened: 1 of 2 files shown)


deploy-edgeX.sh
```diff
@@ -77,7 +77,7 @@ run_service export-distro
 
 run_service app-service-rules
 
-run_service rulesengine
+[ "$(uname -m)" = "x86_64" ] && run_service rulesengine
```
cloudxxx8, line 80: please also remove line 94
sleep 100s # Wait for rulesengine fully startup, because it takes around 100s on Raspberry Pi

jinlinGuan: removed

cloudxxx8 (approved):
LGTM

PR #374, opened 2020-01-09
Run twice blackbox with Mongo and Redis resp.

### LINKED ISSUE #344: Run blackbox twice for both Mongo and Redis database
Blackbox tests should verify both Mongo and Redis database.

jinlinGuan:
Fix https://github.com/edgexfoundry/blackbox-testing/issues/344

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,78 @@
+pipeline {
+    agent { label "${env.SLAVES}" }
+    options {
+        timeout(time: 2, unit: 'HOURS')
+        timestamps()
+    }
+    stages {
+        stage("BuildAndTest") {
+            matrix {
+                agent { label "${env.SLAVES}" }
+                environment {
+                    MACHINE = sh(script: 'uname -m', returnStdout: true).trim()
+                }
+                axes {
+                    axis {
+                        name "DATABASE"
+                        values "mongo", "redis"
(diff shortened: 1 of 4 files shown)


jinlinGuan:
wait for [edgexfoundry/developer-scripts#208](https://github.com/edgexfoundry/developer-scripts/pull/208) to be merged

bnevis-i:
I don't like the way this is done.  There is a lot of one-time initialization with the security services and restating the test suite with a different set of containers might yield a different result than running the entire suite on a fresh VM.  I'd rather CI make a redis version of the blackbox-testing jobs.

cloudxxx8:
@bnevis-i we understand your concern, but adding Jenkins jobs might not be a good solution.  The current number of Jenkins jobs for one release is four, and it would become eight.  After implementing the compatibility tests might become four times.
@cherrycl has proposed a better solution.  we could leverage the parallel run feature of Jenkins to trigger those tests on different virtual machines and combine the report together.
we will update this PR after confirming the approach.

jinlinGuan:
updated, using `redis` as default and adding a jenkinsfile that can run the tests with Mongo and Redis in parallel.

jinlinGuan:
Closing this PR. May use Jenkins Pipeline in the future, but still use JJB for now.

PR #385, opened 2020-02-05
Update to use Redis by default

### LINKED ISSUE #383: Change default persistence to Redis
Update scripts to default to Redis

andresrinivasan:
Fixes #383.

Tested via blackbox tests using the 1.1.0 Docker containers

Also

- _Chmod'ed +x_ a couple of files in `bin` that were missing the execute bit
- Tweaked `bin/run.sh` so that it is independent of `$(cwd)`
- Added support for referencing a Docker Compose file that was on the file system
- Updated docs, tried to simplify some of the steps, removed requirement to edit scripts


jinlinGuan:
There are some relevant changes in [edgexfoundry/blackbox-testing/374](https://github.com/edgexfoundry/blackbox-testing/pull/374)

sync.sh
```diff
@@ -7,14 +7,18 @@
 [ "$SECURITY_SERVICE_NEEDED" != true ] && USE_NO_SECURITY="-no-secty"
 
 # redis or mongo
-[ "$FOR_REDIS" = true ] && USE_REDIS="-redis"
+if [ "${FOR_REDIS:=true}" = true ]; then 
+     PERSIST="-redis"
+else
+     PERSIST="-mongo"
```
jinlinGuan, line 13: The names of the compose files in developer-scripts don't contain "mongo".

bin/run.sh
```diff
@@ -8,15 +8,10 @@ fi
 
 # Ensure we fail the job if any steps fail
 set -e -o pipefail
-#. $(dirname "$0")/env.sh
 
 # Run the compose file for blackbox testing
 export network=$(docker network ls | awk '{print $2}' | grep edgex-network)
-docker-compose -f ../docker-compose-test-tools.yml up -d app-service-configurable
-
-option="${1}"
```
jinlinGuan, line 17: `option` is needed for the switch case.

bin/run.sh
```diff
@@ -8,15 +8,10 @@ fi
...
-
-echo " ${option}"
+docker-compose -f "$(dirname "$0")"/../docker-compose-test-tools.yml up -d app-service-configurable
```
jinlinGuan, line 16: The `$(dirname "$0")` is `bin`, which causes ERROR here. The `docker-compose-test-tools.yml` is under the parent directory.

jinlinGuan (changes requested):

sync.sh
andresrinivasan: Please see https://github.com/edgexfoundry/developer-scripts/pull/218

jinlinGuan: Then the compose files with security services (docker-compose-nexus.yml and docker-compose-nexus-arm64.yml) need to contain "mongo" as well.

jinlinGuan: PR [edgexfoundry/developer-scripts#223](https://github.com/edgexfoundry/developer-scripts/pull/223) fixed it

jinlinGuan:
@andresrinivasan This PR need to be merged ASAP in order to run tests on Jenkins properly, since [edgexfoundry/developer-scripts#218](https://github.com/edgexfoundry/developer-scripts/pull/218) has been merged. Thanks.

bin/run.sh
andresrinivasan: Apologies for the fat finger. Corrected.

bin/run.sh
andresrinivasan: What should be happening is `bin/run.sh` and `cd bin && ./run.sh` should now both work. In the former case `$(dirname $0)` will be `bin` so `bin/../docker-....yml` should be found. In the latter case, `$(dirname $0)` will be `.` so `./../docker-....yml` should be found. I'm seeing this in my testing.

How do I look at the failure in the Jenkins build?

andresrinivasan: The issue was not line 14 but line 147. The `updateMaxResultCount.sh` script didn't have an exit statement and the `set -e` on line 12 was triggered. This is corrected now.

bin/run.sh
```diff
@@ -134,7 +131,6 @@ testAll() {
   coreDataTest
   commandTest
   metaDataTest
-  loggingTest
```
jinlinGuan, line 137: Is that because Redis doesn't need `loggingTest`? Mongo still needs it, so I don't think  it should be removed now.

cloudxxx8: @andresrinivasan although we run logging with file as persistence, we still need to run logging test to make sure the API works.
Theoretically, logging service APIs can work with file

bin/run.sh
jinlinGuan: When executing `bin/run.sh -cd`, I got the following ERROR:
image
In this case, `$(dirname $0)` should be added to all the test scripts.

`cd bin && ./run.sh -cd`works fine.

cloudxxx8 (approved):
This PR is mostly good enough.
However, the blackbox test is broken and need this PR to be merged as soon as possible.
I am merging it and Ginny will open other issues to fix the error in this PR.

bin/run.sh
andresrinivasan: Ah. I see what you're seeing now. Opening another issue to track.

PR #397, opened 2020-02-25
Update FOR_REDIS value for security jenkins jobs

### LINKED ISSUE #396: Run Mongo for Security jenkins jobs
There is no redis security-enabled docker-compose, so currently mongo should be used for the security jenkins jobs. But because of the default value `FOR_REDIS` in `deploy-edgeX.sh`, mongo isn't running.
```
if [ "${FOR_REDIS:=true}" = true ]; then
	run_service redis
	persist=redis
else
	run_service mongo
	persist=mongo
fi
```

jinlinGuan:
Fix #396

[Workaround] run mongo database for security enabled test

### DIFF
--- a/deploy-edgeX.sh
+++ b/deploy-edgeX.sh
@@ -74,6 +74,11 @@ if [ "$SECURITY_SERVICE_NEEDED" = "true" ]; then
 	run_service edgex-proxy
 fi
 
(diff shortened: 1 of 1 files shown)


cloudxxx8 (approved):
LGTM

PR #405, opened 2020-03-10
Update to use DATABASE instead of FOR_REDIS

### LINKED ISSUE #404: Change FOR_REDIS to an unambiguoust variable name
Currently, FOR_REDIS=true is using redis db while FOR_REDIS=false is using mongo db. A clearer name will be helpful.

jinlinGuan:
Fix #404
The value is changed from true/false to redis/mongo.


cloudxxx8 (approved):
LGTM

PR #415, opened 2020-03-24
Add test for invalid values in new Reading properties

### LINKED ISSUE #408: Add test for validity of new Reading "data type" properties
New properties to indicate the data type of a Reading.value have been added with the following PR

https://github.com/edgexfoundry/go-mod-core-contracts/pull/223

Please add a test to ensure the proper validity checks are running and a `400` status code is returned from the API in the case of invalid population.

jdharms:
Fixes #408.  Adds requests to cover two cases:

* valueType is declared as Float32 but no floatEncoding is provided
* binaryValue is given but no mediaType is provided

Both of these cases should fail with HTTP 400 errors.

Signed-off-by: Daniel Harms <jdharms@gmail.com>

### DIFF
--- a/bin/postman-test/collections/core-data.postman_collection.json
+++ b/bin/postman-test/collections/core-data.postman_collection.json
@@ -1,6 +1,6 @@
 {
 	"info": {
-		"_postman_id": "c47bc875-a293-4f6d-8cea-639bd82f647a",
+		"_postman_id": "986f05e6-4ffc-40c5-8fbb-59e19bdfd847",
 		"name": "core-data",
 		"schema": "https://schema.getpostman.com/json/collection/v2.1.0/collection.json"
 	},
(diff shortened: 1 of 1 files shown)


AnthonyMBonafide:
LGTM! I verified the new postman requests by importing them into Postman and executing them. However, I would like someone from the QA team to give this a review before I approve. I have a specific setup for my local Linux machine and not sure it is aligned with how the QA environment expects everything to be.

AnthonyMBonafide (commented):
@cloudxxx8 @FelixTing , would you be able to verify these changes?

cloudxxx8:
@jinlinGuan please help review this PR

cloudxxx8:
@jinlinGuan please help review this PR

jinlinGuan:
The testcases pass, but could you please add positive cases for the valid values?

In addition, according to [edgexfoundry/go-mod-core-contracts#223](https://github.com/edgexfoundry/go-mod-core-contracts/pull/223/files#diff-1a2f491a07674de4376b10f16388c3aeR124-R126), missing mediaType should have the Error messge in edgex-core-data log: "media type must be specified for binary values", but I got "Error decoding the reading: illegal base64 data at input byte 2" whether mediaType is given or not.

jdharms:
@jinlinGuan Good catch on the trivial failure case.  The fake binary data I was using was invalid and that was tripping the validator.  I've replaced it with valid CBOR and now the correct error message is coming through on my end.

The positive cases are covered in another PR of mine: https://github.com/edgexfoundry/blackbox-testing/pull/416

I created two because there were two issues assigned to me.  I can collapse them into one PR if that is preferred.

Thank you for the review, much appreciated.

jinlinGuan (approved):
Verified with nightly-build redis no security compose file locally.

AnthonyMBonafide (approved):
LGTM! and also verified by @jinlinGuan

PR #428, opened 2020-04-14
Remove redundant port in the event testcase

### LINKED ISSUE #427: Redundant port in Core-data evnet testcase
In testcase `http://localhost:48080/api/v1/event event can be added with binary value`, `{{baseUrl}}` includes port, so the `port` section should be removed.
```
"url": {
    "raw": "{{baseUrl}}/api/v1/event",
    "host": [
        "{{baseUrl}}"
    ],
    "port": "48080",
    "path": [
        "api",
        "v1",
        "event"
    ]
}
```

jinlinGuan:
Fix https://github.com/edgexfoundry/blackbox-testing/issues/427


cloudxxx8 (approved):
LGTM

PR #431, opened 2020-04-18
Added test cases to cover common provision watcher errors

### LINKED ISSUE #430: Provision Watcher endpoints do not have basic coverage
Provision Watcher endpoints, especially the error endpoints, do not have full coverage for all scenarios defined in the RAML. This will be updated with new scenarios added.

brandonforster:
Fixes #430 

This does not cover every error case (there's lots of 500 catch-alls that are basically impossible to get)  but this should be a marked increase in the amount of coverage we have.

One thing I would like to point out: current behavior is to treat IDs that do not conform to either the BSON or UUID spec as "correct, but nonexistent". Redis doesn't care what the ID is or if it's in no format at all and handles this without issue, but the Mongo driver does and needs a valid format. I know we just merged an issue trying to address some of this in edgex-go, but being opinionated about ID format and returning a 400 instead of a 404 on an invalid ID format might be a topic that we explore for Hanoi.

Signed-off-by: Brandon Forster <me@brandonforster.com>

### DIFF
--- a/bin/postman-test/collections/core-metadata.postman_collection.json
+++ b/bin/postman-test/collections/core-metadata.postman_collection.json
@@ -9067,7 +9067,54 @@
 								"{{invalidProvisionWatcherId}}"
 							]
 						},
-						"description": "Remove the ProvisionWatcher designated by the database generated id for the ProvisionWatcher. Returns ServiceException (HTTP 503) for unknown or unanticipated issues. Returns NotFoundException (HTTP 404) if no provision watcher with the provided id is found."
+						"description": "Attempt to delete a provision watcher with an incorrectly formatted ID."
+					},
+					"response": []
+				},
+				{
+					"name": "309 http://localhost:48081/api/v1/provisionwatcher/id/:id",
+					"event": [
+						{
+							"listen": "test",
+							"script": {
+								"type": "text/javascript",
+								"exec": [
+									"/*",
+									" * Test Case:  /provisionwatcher/id/:id - 404 Error - DEL ",
+									" * Version: Alpha",
+									" * @Author: Brandon Forster",
+									" *",
+									" */",
+									" ",
+									"//Test Case for status : 404- valid, non existent ID",
+									"tests[\"Status code is 404\"] = responseCode.code === 404;",
+									"//Test response time",
+									"tests[\"Response time is less than \"+data.responseTime+\"ms\"] = responseTime < data.responseTime;"
+								]
+							}
+						}
+					],
+					"request": {
+						"method": "DELETE",
+						"header": [],
+						"body": {
+							"mode": "raw",
+							"raw": ""
+						},
+						"url": {
+							"raw": "{{baseUrl}}/api/v1/provisionwatcher/id/{{validId}}",
+							"host": [
+								"{{baseUrl}}"
+							],
+							"path": [
+								"api",
+								"v1",
+								"provisionwatcher",
+								"id",
+								"{{validId}}"
+							]
+						},
+						"description": "Attempt to delete a provision watcher with a valid (but nonexistent) UUID"
 					},
 					"response": []
 				},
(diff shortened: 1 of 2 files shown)


bin/postman-test/collections/core-metadata.postman_collection.json
```diff
@@ -9388,7 +9656,544 @@
 								"provisionwatcher"
 							]
 						},
-						"description": "Add a new ProvisionWatcher - name must be unique. Returns ServiceException (HTTP 503) for unknown or unanticipated issues.  Returns DataValidationException (HTTP 409) if profile service are unknown"
+						"description": "Attempt to update a provision watcher using an invalid request body (in this case, an addressable)."
+					},
+					"response": []
+				},
+				{
+					"name": "316 http://localhost:48081/api/v1/provisionwatcher",
```
jinlinGuan, line 9664: I got `404 Not Found` for this testcase.

brandonforster: Can you ensure you have the latest from master on edgex-go? I ran these again this morning and I'm seeing the 409, as expected.

cloudxxx8: @brandonforster Ginny might test it before the latest docker images ready.  we will check again tomorrow.

jinlinGuan: I still got `404 Not Found`.

I used the following steps:
1. `bash deploy-edgeX.sh`
2.  `. bin/env.sh`
3. `bash bin/run.sh -md`

brandonforster: Found the issue, I had some lingering test data in my local instance (not Docker.) Should work now, thanks for the catch!

bin/postman-test/collections/core-metadata.postman_collection.json
```diff
@@ -9388,7 +9656,522 @@
...
+								"v1",
+								"provisionwatcher"
+							]
+						},
+						"description": "Attempt to update a provision watcher using a malformed ID."
+					},
+					"response": []
+				},
+				{
+					"name": "306 http://localhost:48081/api/v1/provisionwatcher",
```
brandonforster, line 9815: This is where 306 ended up. It should be unchanged.

bin/postman-test/collections/core-metadata.postman_collection.json
```diff
@@ -9388,7 +9656,522 @@
...
+								"provisionwatcher"
+							]
+						},
+						"description": "Attempt to add a provision watcher using an invalid JSON request body, not specifying a name"
+					},
+					"response": []
+				},
+				{
+					"name": "320 http://localhost:48081/api/v1/provisionwatcher/:id",
+					"event": [
```
jinlinGuan, line 9966: This testcase returned `500 Internal Server Error` instead of `404` when I tested it with mongo no security docker compose.

bin/postman-test/collections/core-metadata.postman_collection.json
jinlinGuan: It returns the correct status code now.

bin/postman-test/collections/core-metadata.postman_collection.json
brandonforster: That's an issue with the REST logic in EdgeX that this is exposing, not the test.

@AnthonyMBonafide I'll create a follow up, I must have missed this in our first pass.

brandonforster: https://github.com/edgexfoundry/edgex-go/pull/2512 should address this problem on EdgeX-Go. I believe we can merge this PR as-is.

jinlinGuan (approved):
LGTM

cloudxxx8 (approved):
LGTM

PR #446, opened 2020-05-04
Add testcase with logner timeout to core-command

### LINKED ISSUE #445: Add Get core-command device test with 45 seconds timeout
The first call is expected to take more time, so add a testcase with longer timeout for the first call, and keep the original testcase for the second call.

jinlinGuan:
Fix https://github.com/edgexfoundry/blackbox-testing/issues/445

related: https://github.com/edgexfoundry/edgex-go/issues/2526


cloudxxx8 (approved):
LGTM

PR #448, opened 2020-05-04
refactor: Remove settings no longer needed on app-service-configurable

### LINKED ISSUE #447: Cleanup App Service Configurable overrides in docker-compose-test-tools.yml
Latest App Service Configurable does not require as many overrides.

No longer need to skip version check. It handles master as the version from Core Data
No longer need to override edgex_registry. This is default in command line for Config Provider

lenny-goodell:
No longer need to skip version check. It handles master as the version from Core Data
No longer need to override edgex_registry. This is default in command line for Config Provider
Do have to add Registry_Host: edgex-core-consul once is removed

closes #447

### DIFF
--- a/docker-compose-test-tools.yml
+++ b/docker-compose-test-tools.yml
@@ -28,16 +28,15 @@ services:
     image: ${appService}
     ports:
       - "48095:48095"
-    entrypoint: ["/app-service-configurable"]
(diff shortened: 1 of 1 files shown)


jinlinGuan (approved):
LGTM

cloudxxx8 (approved):
LGTM

PR #471, opened 2020-10-15
feat: blackbox tests running against docker or native (snap)

hareshkainthcanonical:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://wiki.edgexfoundry.org/display/FA/Contributor%27s+Guide
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [x] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?

Blackbox tests would run exclusively with docker containers.

Issue Number:

## What is the new behavior?
Blackbox tests can now run against either docker or snaps.

## Does this PR introduce a breaking change?

- [ ] Yes
- [X] No

## Are there any specific instructions or things that should be known prior to reviewing?
# Blackbox Testing against Edgexfoundry Snap

Helpers scripts have been created to support blackbox testing against Edgexfoundry snap. The scripts will install the appropriate snaps and enable required services for testing.

Once the helper scripts have been executed the tester must then run two additional commands that will execute the actual blackbox tests.

Step 1: Setup environment variable

Locate the env.sh file at the following path

**/bin/env.sh**

Uncomment the following environment variable

**export TEST_ENV=localhost**

If you wish to use different channels for edgexfoundry and/or app-service-configurable then add the following environment variable

**export EF_CHANNEL=stable**

Accepted values for EF_CHANNEL are stable, edge, candidate

**export ASC_CHANNEL=stable**

Accepted values for ASC_CHANNEL are stable, edge, candidate

You can also use locally built snaps of either edgexfoundry and/or app-service-configurable by specifying the following environment variables

**export EF_SNAP=path_to_snap**

**export ASC_SNAP=path_to_snap**

Replace path_to_snap with an absolute path to each snap respectively.

Note, if both ASC_CHANNEL and ASC_SNAP variables are exported then ASC_SNAP will take precendence and ASC_CHANNEL will be ignored. This is also true for EF_SNAP and EF_CHANNEL.

If you wish to run the tests with security than set the following environment variable:

**export SECURITY_SERVICE_NEEDED=true**

Once you have set the environment variables, save the file. Navigate back to the root directory of the project and run the following bash command

**. bin/env.sh**

Step 2: Deploy test environment

Navigate to the following directory

**cd misc/scripts/**

Run the following script

**./deploy-edgeX-snaps.sh**

The above script will install the appropriate snaps and enable the services needed for testing.

Step 3: Execute blackbox tests:

**./bin/run.sh -all** 

The above **-all** flag executes all tests

Note you can run individual tests using different flags, however, it has been determined that the tests have to run in a specific sequence in order to ensure no assertions fail. The exact sequence is as follows:

**-dv** = deviceVirtualTest

**-asc** = appServiceConfigurableTest

**-cd** = coreDataTest

**-co** = commandTest

**-md** = metaDataTest

**-sn** = supportNotificationTest

**-sys** = supportSchedulerTest

## Other information
Tested against both Snaps and Docker.


bin/cleanSecurityserviceTest.sh
```diff
@@ -1,8 +1,5 @@
 #!/bin/bash
 
 echo "Info: Clean Securityservice's test data."
-
-COMPOSE=$(dirname "$0")/../$(ls $(dirname "$0")/../ | awk '/docker-compose/ && !/test-tools/')
-docker-compose -p edgex -f ${COMPOSE} run --rm --entrypoint /edgex/security-proxy-setup edgex-proxy --init=false --userdel=testuser
-
+sudo edgexfoundry.security-proxy-setup-cmd --init=false --userdel=testuser
```
tonyespy: This looks incorrect to me, as you're removing the docker-specific command and replacing with the snap command? This file should check the testing environment variable and act accordingly.

bin/postman-test/environment/device-virtual.postman_environment.json
```diff
@@ -1,7 +1,17 @@
 {
-	"id": "e86f1c07-db3f-4ec2-ba52-5c32f2ec2865",
+	"id": "e86f1c07-db3f-4ec2-ba52-5c32f2ec2866",
```
tonyespy: Why did you change `id`?

bin/postman-test/environment/device-virtual.postman_environment.json
```diff
@@ -19,15 +29,10 @@
...
-			"value": ":49990",
-			"description": "",
-			"enabled": true
 		}
 	],
 	"_postman_variable_scope": "environment",
-	"_postman_exported_at": "2019-05-06T16:17:46.914Z",
-	"_postman_exported_using": "Postman/6.7.4"
+	"_postman_exported_at": "2020-01-02T10:06:25.044Z",
+	"_postman_exported_using": "Postman/7.13.0"
```
tonyespy: Did you actually use postman to re-export this file? Otherwise I would leave these fields untouched.

bin/postman-test/environment/device-virtual.postman_environment.json
```diff
@@ -19,15 +29,10 @@
...
 }
+
```
tonyespy: Was this edit to fix the lack of a trailing newline?

bin/run.sh
```diff
@@ -3,16 +3,55 @@
 # Set security environment variable. This is mostly cosmetic as the tests will display warnings that
 # the env is not set.
 if [ "$SECURITY_SERVICE_NEEDED" = "true" ]; then
+  echo "In security mode: true"
```
tonyespy: I don't think is necessary to enable snap testing, so it should either be in a separate commit that increases the verbosity of the script output or dropped altogether.

bin/run.sh
```diff
@@ -3,16 +3,55 @@
...
 	export SECURITY_IS_ON="true"
 else
+  echo "In security mode: false"
```
tonyespy: Same as the previous comment...

bin/run.sh
```diff
@@ -190,7 +229,7 @@ case $1 in
 	testAll | tee $EDGEXLOGSPATH
 	;;
 *)
-	echo "$(basename ${0}):usage: [-cd Coredata] | [-md Metadata] | [-co Command] | [-sn SupportNotification] | [-ss SupportScheduler] | [-dv DeviceVirtual] | [-asc AppServiceConfigurable] | [-sec securityTest] | [-all All]"
+	echo "$(basename "${0}"):usage: [docker Coredata] | [-md Metadata] | [-co Command] | [-sn SupportNotification] | [-ss SupportScheduler] | [-dv DeviceVirtual] | [-asc AppServiceConfigurable] | [-sec securityTest] | [-all All]"
```
tonyespy: Did you really mean to change `-cd` to `docker` here? 

Also are the additional quotes the result of a shellcheck warning? In general, you should always use shellcheck for new scripts you add, or new statements you add to existing scripts, but if you're modifying lines in an existing scripts that have nothing to do with your PR, those changes belong in a separate commit.

bin/run.sh
```diff
@@ -200,4 +239,4 @@ echo
 echo "Info: Logs available in [scriptLogs]"
 #echo "Info: HTML Reports available in [Reports]"
 echo
-sh $(dirname "$0")/endBanner.sh
+sh "$(dirname "$0")/endBanner.sh"
```
tonyespy: Same comment as above re: shellcheck and existing scripts.

bin/security/setupSecurityAccount.sh
```diff
@@ -2,23 +2,20 @@
...
   -useradd)
   echo "Info: Add security account."
-  OT=$(docker-compose -p edgex -f ${COMPOSE} run --rm \
-    --entrypoint /edgex/security-proxy-setup edgex-proxy --init=false --useradd=testuser --group=admin | grep '^the access token for')
-  TOKEN=$(echo ${OT} | sed 's/.*: \([^.]*\.[^.]*\.[^.]*\).*/\1/')
-  export Token=${TOKEN}
-  echo ${TOKEN}
-  ;;
 
+  OT=$(sudo edgexfoundry.security-proxy-setup-cmd --confdir=/var/snap/edgexfoundry/current/config/security-proxy-setup/res --init=false --useradd=testuser --group=admin | grep '^the access token for')
```
tonyespy: As with one of the other security scripts, you've changed this script to be snap-only which will fail if/when used to run the tests against docker-deployed edgex.

bin/security/setupSecurityAccount.sh
```diff
@@ -2,23 +2,20 @@
...
+  TOKEN=$(echo $OT | sed 's/.*: \([^.]*\.[^.]*\.[^.]*\).*/\1/')
```
tonyespy: This line generates shellcheck warnings.

Also just curious what the value of `$OT` is and whether you could use a simpler sed expression? If not, then please at least add a comment explaining the expression you've used here because even I would need to go look this up to figure out what it's doing.

bin/security/setupSecurityAccount.sh
```diff
@@ -2,23 +2,20 @@
...
+  export TOKEN
+
+  echo "Generated token: " $TOKEN
```
tonyespy: shellcheck warning

bin/security/setupSecurityAccount.sh
```diff
@@ -2,23 +2,20 @@
...
+
+  ;;
   -userdel)
   echo "Info: Delete security account."
-  docker-compose -p edgex -f ${COMPOSE} run --rm \
-    --entrypoint /edgex/security-proxy-setup edgex-proxy --init=false --userdel=testuser
+  sudo edgexfoundry.security-proxy-setup-cmd --init=false --userdel=testuser
```
tonyespy: See above comment about the fact that you've now made this script snap-specific.

bin/setupSecurityserviceTest.sh
```diff
@@ -1,13 +1,16 @@
...
+TOKEN=$(echo $OT | sed 's/.*: \([^.]*\.[^.]*\.[^.]*\).*/\1/')
+export TOKEN
+echo "Generated token: " $TOKEN
+
+RT=$(sudo sh -c "cat /var/snap/edgexfoundry/current/secrets/edgex-security-proxy-setup/secrets-token.json")
+VAULTKEY=$(echo $RT | sed 's/.*"client_token":"\([^"]*\)".*/\1/')
+export VAULTKEY
+
+echo VAULT_KEY: $VAULTKEY
+echo "Generated vault key: " VAULT_KEY
```
tonyespy: Same comments as above:

  * this script is now snap-specific
  * shellcheck warnings present

docker-compose-test-tools.yml
```diff
@@ -50,6 +50,7 @@ services:
 
   postman:
     image: ${postman}
+#    network_mode: "host"
```
tonyespy: It'd be nicer if we didn't add a commented out line to this file and instead used sed's capability to append a line to a file. If that's not possible, then at least explain an additional comment line that explains why this present, but I'm pretty sure you can make this work without this line.

docker-compose-test-tools.yml
```diff
@@ -64,6 +65,4 @@ services:
 networks:
   edgex-network:
     external:
-     name: edgex_edgex-network
```
tonyespy, line 67: As mentioned previously, if you're going to clean up formatting/whitespace, please split out into a separate commit. In general, your PR should only include changes which are necessary to enable the bug fix and/or feature that you're contributing.

docs/How-to-run-blackbox-testing-against-snaps.md
```diff
@@ -0,0 +1,79 @@
+# Blackbox Testing against Edgexfoundry Snap
+
```
tonyespy, line 2: I would start with something like this:

This file contains instructions for how to run the blackbox-tests against the EdgeX Foundry snaps edgexfoundry and edgex-app-service-configurable that are both available from the Snap Store (https://snapcraft.io/store).

docs/How-to-run-blackbox-testing-against-snaps.md
```diff
@@ -0,0 +1,79 @@
...
+Helpers scripts have been created to support blackbox testing against Edgexfoundry snap. The scripts will install the appropriate snaps and enable required services for testing.
+
+Once the helper scripts have been executed the tester must then run two additional commands that will execute the actual blackbox tests.
+
```
tonyespy: I would drop the above three sentences, as they aren't really adding value here... The instructions below should be sufficient.

misc/scripts/addSecurityAccountTestUser.sh
```diff
@@ -0,0 +1,11 @@
...
+
+echo "Adding security account for testuser..."
+
+TEST_USER_ACCESS_TOKEN=$(sudo edgexfoundry.security-proxy-setup-cmd --confdir=/var/snap/edgexfoundry/current/config/security-proxy-setup/res --init=false --useradd=testuser --group=admin | grep '^the access token for')
+
+export TEST_USER_ACCESS_TOKEN
+
+echo "TEST_USER_ACCESS_TOKEN = " $TEST_USER_ACCESS_TOKEN
+
+echo "Adding security account for testuser...Done"
```
tonyespy: I'm not sure why this script is necessary as there are two other scripts which do the same thing (and you've already modified both)?

 * `bin/security/setupSecurityAccount.sh`
 * `bin/setupSecurityserviceTest.sh`

misc/scripts/teardownSecurityUser.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+echo "Removing security testuser"
+
+. removeSecurityAccountTestUser.sh
+
+echo "Removing security testuser...Done"
```
tonyespy: I'm not sure why you needed to add this script, as it just delegates to `removeSecurityAccountTestUser.sh`, and only adds some additional output which echoes the output of script it calls?

  * "Removing security testuser" vs. "Removing security account..."
  * "Removing security testuser...Done" vs. "Removing security account...Done"

misc/scripts/removeSecurityAccountTestUser.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+echo "Removing security account..."
+
+sudo edgexfoundry.security-proxy-setup-cmd --confdir=/var/snap/edgexfoundry/current/config/security-proxy-setup/res --init=false --userdel=testuser
+
+echo "Removing security account...Done"
```
tonyespy: I'm not sure why this script is necessary as there are two other scripts which do the same thing (and you've already modified both)?

  * `bin/cleanSecurityserviceTest.sh`
  * `bin/security/setupSecurityAccount.sh`

misc/scripts/deploy-edgeX-snaps.sh
```diff
@@ -0,0 +1,169 @@
...
+sudo snap set edgexfoundry device-virtual=on
+
+echo "$(timestamp) Setting edgex-app-service-configurable profile to blackbox-tests..."
+sudo snap set edgex-app-service-configurable profile=blackbox-tests
+
+echo "$(timestamp) Starting edgex-app-service-configurable service..."
+sudo snap start edgex-app-service-configurable.app-service-configurable
+
+echo "$(timestamp) Generating security token. Token is used for tests when in secure mode..."
+TOKEN=$(echo "$TEST_USER_ACCESS_TOKEN" | sed 's/.*: \([^.]*\.[^.]*\.[^.]*\).*/\1/')
```
tonyespy: shellcheck warning:

        ^-- SC2001: See if you can use ${variable//search/replace} instead.

Also per my earlier comment about the sed replace pattern being used here, can this be simplified? If not, please add a comment which explains the purpose of this expression.

docs/How-to-run-blackbox-testing-against-snaps.md
```diff
@@ -0,0 +1,79 @@
...
+Uncomment the following environment variable
+
+**export TEST_ENV=localhost**
+
+If you wish to use different channels for edgexfoundry and/or app-service-configurable then add the following environment variable
+
+**export EF_CHANNEL=stable**
+
+Accepted values for EF_CHANNEL are stable, edge, candidate
```
tonyespy: Actually accepted channels are: `[track/]<risk>[/branch]`, where <risk> can be beta, candidate, edge, or stable.

docs/How-to-run-blackbox-testing-against-snaps.md
```diff
@@ -0,0 +1,79 @@
...
+
+**export ASC_CHANNEL=stable**
+
+Accepted values for ASC_CHANNEL are stable, edge, candidate
```
tonyespy: Same as above...

docs/How-to-run-blackbox-testing-against-snaps.md
```diff
@@ -0,0 +1,79 @@
...
+
+Once you have set the environment variables, save the file. Navigate back to the root directory of the project and run the following bash command
+
+**. bin/env.sh**
+
+Step 2: Deploy test environment
+
+Navigate to the following directory
+
+**cd misc/scripts/**
```
tonyespy: Unless @cloudxxx8 or @cherrycl suggested moving this script to misc/scripts, why not put at the top-level like the other deploy script? I'll defer to their opinion on this though...

bin/run.sh
```diff
@@ -3,16 +3,55 @@
...
 
-# Run the compose file for blackbox testing
-docker-compose -f ${docker_compose_test_tools} up -d app-service-configurable
+# Run the compose file for blackbox testing (only in docker)
+ENV_DOCKER_OR_SNAP_FLAG=""
+
+test_with_docker() {
+  echo "Testing against docker"
+  network=$(docker network ls | awk '{print $2}' | grep edgex-network)
+  export network
```
tonyespy: I couldn't find any other references to an environment variable called `network`. What is this supposed to accomplish?

Also FYI, environment variables are usually uppercase only.

bin/run.sh
```diff
@@ -3,16 +3,55 @@
...
 # Ensure we fail the job if any steps fail
 set -o pipefail
 
-# Run the compose file for blackbox testing
-docker-compose -f ${docker_compose_test_tools} up -d app-service-configurable
+# Run the compose file for blackbox testing (only in docker)
+ENV_DOCKER_OR_SNAP_FLAG=""
+
+test_with_docker() {
+  echo "Testing against docker"
```
tonyespy: I wouldn't add any additional output to when testing with docker as this has the potential to break existing automation tools used by the project that might not be expecting this additional output. Again, "less is more".

bin/run.sh
```diff
@@ -3,16 +3,55 @@
...
+  network=$(docker network ls | awk '{print $2}' | grep edgex-network)
+  export network
+
+  sed -i '32,49s/^#//' "$docker_compose_test_tools"
+  sed -i '54,55s/^#//' "$docker_compose_test_tools"
+  sed -i '65,68s/^#//' "$docker_compose_test_tools"
+  sed -i '53,53s/^ /#/' "$docker_compose_test_tools"
```
tonyespy: These looks super-fragile. What happens if additional lines are added to the docker-compose file? In general when using sed, pattern matching (or addressing) is a much safer approach.

Also while what you've done here is clever, I would instead just delete the lines from the files when running in localhost mode, and not worry about re-adding the lines when running against docker. It's doubtful someone would checkout blackbox-tests, run against snaps, and then run against docker. If someone did want to do this, they could also easily just run `git checkout -- <name of compose file>` to restore the edits.

bin/run.sh
```diff
@@ -3,16 +3,55 @@
...
 	export SECURITY_IS_ON="false"
 fi
 
 # Ensure we fail the job if any steps fail
 set -o pipefail
 
-# Run the compose file for blackbox testing
-docker-compose -f ${docker_compose_test_tools} up -d app-service-configurable
+# Run the compose file for blackbox testing (only in docker)
+ENV_DOCKER_OR_SNAP_FLAG=""
```
tonyespy: Since `test_with_docker` and `test_with_localhost` both set this variable, there's no need to initialize it to "" here.

bin/run.sh
```diff
@@ -3,16 +3,55 @@
...
+  echo "In security mode: false"
 	export SECURITY_IS_ON="false"
 fi
 
 # Ensure we fail the job if any steps fail
 set -o pipefail
 
-# Run the compose file for blackbox testing
-docker-compose -f ${docker_compose_test_tools} up -d app-service-configurable
+# Run the compose file for blackbox testing (only in docker)
```
tonyespy: This comment should be moved into `test_with_docker`.

bin/run.sh
```diff
@@ -3,16 +3,55 @@
...
+
+if [[ -n "$TEST_ENV" ]]; then
+  case $TEST_ENV in
+    docker) test_with_docker ;;
+    localhost) test_with_localhost ;;
+    *) echo "$(basename "${0}") - TEST_ENV usage: docker | localhost" && echo "$(basename "${0}") - TEST_ENV is currently set to $TEST_ENV" && exit 0
+  esac
+else
+  test_with_docker
+fi
```
tonyespy, line 40: I would just simplify this to:

```
if [[ "$TEST_ENV" == "localhost"]]; then
  test_with_localhost
else
  test_with_docker
```

It's doubtful anyone is actually going to explicitly set `TEST_ENV=docker`, or anything else...

bin/run.sh
```diff
@@ -3,16 +3,55 @@
...
+
+test_with_localhost() {
+  echo "Testing against localhost"
+
+  sed -i '32,49s/^/#/' "$docker_compose_test_tools"
+  sed -i '54,55s/^/#/' "$docker_compose_test_tools"
+  sed -i '65,68s/^/#/' "$docker_compose_test_tools"
+  sed -i '53,53s/^#//' "$docker_compose_test_tools"
+
+  ENV_DOCKER_OR_SNAP_FLAG=""
```
tonyespy: This really isn't a flag, it's an optional suffix which is only ever set to `-docker` or `""`, so I would shorten this to **ENV_SUFFIX**. 

Also instead of exporting the variable later in the script, why not just do it here (and same in `test_with_docker`)?

bin/env.sh
```diff
@@ -4,3 +4,27 @@ export appService=nexus3.edgexfoundry.org:10004/docker-app-service-configurable:
...
+# Note, EF_SNAP takes precedence over EF_CHANNEL and ASC_SNAP takes precedence over ASC_CHANNEL.
+# If EF_CHANNEL is not set or not a valid value, EF_CHANNEL will default to stable.
+# Accepted values are: stable | beta | edge | candidate
+#export EF_CHANNEL=stable
+
+# If ASC_CHANNEL is not set or not a valid value, ASC_CHANNEL will default to stable.
+# Accepted values are: stable | beta | edge | candidate
+#export ASC_CHANNEL=edge
+
+# If EF_SNAP is not set or not a valid value, deploy-edgeX-snaps script will exit.
```
tonyespy: This shouldn't be true, if I set **EF_CHANNEL** then I'm asking for the snap to be installed from the store. In that case **EF_SNAP** won't be set, and `deploy-edgeX-snaps.sh` script should still work right?

bin/env.sh
```diff
@@ -4,3 +4,27 @@ export appService=nexus3.edgexfoundry.org:10004/docker-app-service-configurable:
...
+# Set test environment to either docker or localhost (snaps).
+# If TEST_ENV is not set then docker will be used as default.
+# Accepted values are: docker | localhost
+#export TEST_ENV=localhost
+
+# Set any of the following snap environment variables. Note you must also set TEST_ENV=localhost as well
+# Note, EF_SNAP takes precedence over EF_CHANNEL and ASC_SNAP takes precedence over ASC_CHANNEL.
+# If EF_CHANNEL is not set or not a valid value, EF_CHANNEL will default to stable.
+# Accepted values are: stable | beta | edge | candidate
+#export EF_CHANNEL=stable
```
tonyespy: I'm not sure these belong in this script as they're only ever used by our `deploy-edgeX-snaps.sh` script. If you really think they need to be someplace commented out like this, then the deploy script is probably a better place for them.

bin/env.sh
```diff
@@ -4,3 +4,27 @@ export appService=nexus3.edgexfoundry.org:10004/docker-app-service-configurable:
...
+
+# If ASC_CHANNEL is not set or not a valid value, ASC_CHANNEL will default to stable.
```
tonyespy: Except that it actually defaults to **edge** here...

docs/How-to-run-blackbox-testing-against-snaps.md
```diff
@@ -0,0 +1,79 @@
...
+
+Locate the env.sh file at the following path
+
+**/bin/env.sh**
+
+Uncomment the following environment variable
+
+**export TEST_ENV=localhost**
+
+If you wish to use different channels for edgexfoundry and/or app-service-configurable then add the following environment variable
```
tonyespy: **suggest**: ...then define the following environment **variables**.

misc/scripts/deploy-edgeX-snaps.sh
```diff
@@ -0,0 +1,169 @@
...
+timestamp() {
+  local ts=""
+  ts=$(date +"%Y-%m-%dT%T.%3N") # current time
+  echo "$ts -"
+}
+
+channel_type() {
+  local CHANNEL_VAL=""
+
+  if [[ -z "$1" ]]; then
```
tonyespy: I think it's probably overkill to validate the channels, especially given the fact that tracks and/or branches can be specified. I'd just pass the env var through to the snap command, and if it fails, then an error will be output.

misc/scripts/deploy-edgeX-snaps.sh
```diff
@@ -0,0 +1,169 @@
...
+        ASC_SNAP_VAL_VALID=true
+    fi
+fi
+
+echo "============================================================================================"
+echo "                      Deploying edgex snap test service for Blackbox"
+echo "============================================================================================"
+
+echo "$(timestamp) Removing edgexfoundry snap (if already installed)..."
+sudo snap remove edgexfoundry
```
tonyespy: I would also use `--purge` so as to not save a snapshot here.

misc/scripts/deploy-edgeX-snaps.sh
```diff
@@ -0,0 +1,169 @@
...
+        CHANNEL_VAL="--candidate"
+      ;;
+    esac
+  fi
+
+  echo "$CHANNEL_VAL"
+}
+
+EF_CHANNEL_VAL=""
+EF_CHANNEL_VAL_SET=false
```
tonyespy: I also think this logic is a bit overly complex, you could simply do:

```
EF_CHANNEL="${EF_CHANNEL:-stable}"

if [ -n "$EF_SNAP" ]; then
    sudo snap install ${EF_SNAP} --dangerous
elif [ -n "$EF_CHANNEL" ]; then
    sudo snap install edgexfoundry --channel=${EF_CHANNEL}
fi
```

And the same for app-service-configurable...

misc/scripts/deploy-edgeX-snaps.sh
```diff
@@ -0,0 +1,169 @@
...
+
+case $1 in
+--unset-env-var)
+	echo "$(timestamp) Unsetting TEST_ENV, EF_CHANNEL, ASC_CHANNEL, EF_SNAP, ASC_SNAP environment variables after deployment"
+	unset TEST_ENV
+	unset EF_CHANNEL
+	unset ASC_CHANNEL
+  unset EF_SNAP
+  unset ASC_SNAP
+	;;
```
tonyespy: I don't think this is really needed, but if you really think so, please fix the indentation of the last two unset commands.

misc/scripts/deploy-edgeX-snaps.sh
```diff
@@ -0,0 +1,169 @@
...
+esac
+
+echo "============================================================================================"
+echo "              Completed the deployment of edgex snap test service for Blackbox"
```
tonyespy: Note, if someone had specified `--unset-env-var`, then this output doesn't really apply...

tonyespy (changes requested):
Please see comments inline...

Also please refer to the internal google document I shared with you re: the commit message for bulk of your changes.

Your main commit summary says:

_feat: blackbox tests running against docker or native(snap) …_

What your commit does is enable snap testing, so I wouldn't mention docker in commit summary here. Also in general, your commit summary should use the imperative mood:
  * fix xyx
  * cleanup xyx
  * enable foo

My suggestion to you was:

_feat: enable blackbox-tests to be run against snaps_

Your commit body says:

```
Refactored branch to support testing against docker or native(snap). Currently wip. To test against snap then in env.sh file uncomment:

export EF_CHANNEL=stable
```

A couple of points...

 * refactoring means re-implementing code without functional changes
 * this is no longer a "wip" change
 * to enable snap testing you need to set **TEST_ENV** _not_ **EF_CHANNEL**
 * also notice how you need to scroll this page to see your entire commit body, whereas my suggested body below fits onscreen? This is a minor detail, but results in a much easier to look at commit history

My suggestion for this commit was simply:

```
Add support for running the blackbox-tests against EdgeX snaps
or native services. To enable, export the environment variable:

TEST_ENV=localhost
```

docs/How-to-run-blackbox-testing-against-snaps.md
jinlinGuan: I agree with Tony. The `deploy-edgeX-snaps.sh` should be at the top-level. 
The rest of scripts in misc are not necessary since you can modify the `bin/setupSecurityserviceTest.sh`, `bin/cleanSecurityserviceTest.sh`, and `bin/security/setupSecurityAccount.sh` to do the same thing.

jinlinGuan:
There are blackbox-testing automation tests running daily on Jenkins, so please try not to break the current tests running against docker. Thanks.

bin/cleanSecurityserviceTest.sh
hareshkainthcanonical: I forgotten to revert the security changes back to original and then apply the localhost (use snap) changes through an if/else block. Issue resolved.

bin/postman-test/environment/device-virtual.postman_environment.json
hareshkainthcanonical: I didnt manually change anything in this file. I assume when I updated my repo this file also got changed. Will check with upstream and ensure files are in sync.

bin/postman-test/environment/device-virtual.postman_environment.json
hareshkainthcanonical: The docker version (device-virtual-docker.postman_environment) had additional json objects and newer version of postman and timestamp. I updated the localhost version (device-virtual.postman_environment) to the docker version as to bring both in line (sync).

bin/postman-test/environment/device-virtual.postman_environment.json
hareshkainthcanonical: Removed the additional newline.

bin/run.sh
hareshkainthcanonical: Removed.

bin/run.sh
hareshkainthcanonical: Removed.

bin/run.sh
hareshkainthcanonical: Thats my error, it should be -cd. Issue resolved.

bin/security/setupSecurityAccount.sh
hareshkainthcanonical: I forgotten to revert the security changes back to original and then apply the localhost (use snap) changes through an if/else block. Issue resolved.

bin/security/setupSecurityAccount.sh
hareshkainthcanonical: I forgotten to revert the security changes back to original and then apply the localhost (use snap) changes through an if/else block. Issue resolved.

bin/security/setupSecurityAccount.sh
hareshkainthcanonical: I forgotten to revert the security changes back to original and then apply the localhost (use snap) changes through an if/else block. Issue resolved.

bin/setupSecurityserviceTest.sh
hareshkainthcanonical: I forgotten to revert the security changes back to original and then apply the localhost (use snap) changes through an if/else block. Ive now done this.

docs/How-to-run-blackbox-testing-against-snaps.md
hareshkainthcanonical: Ive applied the changes.

docs/How-to-run-blackbox-testing-against-snaps.md
hareshkainthcanonical: Done.

misc/scripts/addSecurityAccountTestUser.sh
hareshkainthcanonical: Will remove this file as its no longer needed.

misc/scripts/teardownSecurityUser.sh
hareshkainthcanonical: This file is no longer needed. Will remove it.

misc/scripts/removeSecurityAccountTestUser.sh
hareshkainthcanonical: This file is no longer needed. Will remove it.

docs/How-to-run-blackbox-testing-against-snaps.md
hareshkainthcanonical: Added beta to description.

docs/How-to-run-blackbox-testing-against-snaps.md
hareshkainthcanonical: Added beta to description.

docs/How-to-run-blackbox-testing-against-snaps.md
hareshkainthcanonical: File moved to top level and all other scripts in misc/scripts directory has been removed. Both misc and scripts directories have also been removed.

bin/run.sh
hareshkainthcanonical: export network no longer used. Removed the export.

bin/run.sh
hareshkainthcanonical: Removed output.

bin/run.sh
hareshkainthcanonical: Removed variable.

hareshkainthcanonical: Removed the variable.

bin/run.sh
hareshkainthcanonical: Comment moved into test_with_docker.

bin/run.sh
hareshkainthcanonical: Replaced with your version.

bin/run.sh
hareshkainthcanonical: Renamed ENV_DOCKER_OR_SNAP_FLAG to ENV_SUFFIX and exported the variable as you suggested.

misc/scripts/deploy-edgeX-snaps.sh
hareshkainthcanonical: Added flag.

bin/env.sh
hareshkainthcanonical: If EF_SNAP is not set then EF_CHANNEL will be used. If EF_CHANNEL is not set then stable version of edgexfoundry will be installed (deploy-edgeX-snaps.sh will continue to work). I need to refine the description to make this point clear.

bin/env.sh
hareshkainthcanonical: Moved these variables to deploy-edgeX.snaps.sh and updated the readme file (How-to-run-blackbox-testing-against-snaps.md).

docs/How-to-run-blackbox-testing-against-snaps.md
hareshkainthcanonical: Refactored the description.

misc/scripts/deploy-edgeX-snaps.sh
hareshkainthcanonical: Removed this code block. Not needed.

misc/scripts/deploy-edgeX-snaps.sh
hareshkainthcanonical: Removed this flag as its not needed.

misc/scripts/deploy-edgeX-snaps.sh
hareshkainthcanonical: Removed the channel validation check.

misc/scripts/deploy-edgeX-snaps.sh
hareshkainthcanonical: Simplified the code for both edgexfoundry and app-service-configurable.

bin/run.sh
hareshkainthcanonical: Now using pattern matching.

bin/run.sh
hareshkainthcanonical: Reverted my change back to the original version.

bin/security/setupSecurityAccount.sh
hareshkainthcanonical: I forgotten to revert the security changes back to original and then apply the localhost (use snap) changes through an if/else block. Issue resolved.

bin/security/setupSecurityAccount.sh
hareshkainthcanonical: shellcheck warning will have to be resolved in another commit.

docker-compose-test-tools.yml
hareshkainthcanonical: Removed the commented code. This has been resolved with regex-type replacement.

docker-compose-test-tools.yml
hareshkainthcanonical: Reverted file back to its original version.

misc/scripts/deploy-edgeX-snaps.sh
hareshkainthcanonical: Removed the security token code as its not needed here.

hareshkainthcanonical:
Ive now addressed the issues you have raised.

bin/env.sh
```diff
@@ -4,3 +4,7 @@ export appService=nexus3.edgexfoundry.org:10004/docker-app-service-configurable:
 export postman=postman/newman:4.5.6
 export docker_compose_test_tools=$PWD/docker-compose-test-tools.yml
 
+# Set test environment to either docker or localhost (snaps).
+# If TEST_ENV is not set then docker will be used as default.
+# Accepted values are: localhost
+#export TEST_ENV=localhost
```
tonyespy: No newline at EOF, did you edit this file on a non-Ubuntu machine?

bin/postman-test/environment/device-virtual.postman_environment.json
tonyespy: If you didn't actually re-run postman to re-generate this file, then as previously suggested, I'd remove these changes to the **postman_exported_*** fields in this file.

bin/run.sh
```diff
@@ -190,7 +209,7 @@ case $1 in
 	testAll | tee $EDGEXLOGSPATH
 	;;
 *)
-	echo "$(basename ${0}):usage: [-cd Coredata] | [-md Metadata] | [-co Command] | [-sn SupportNotification] | [-ss SupportScheduler] | [-dv DeviceVirtual] | [-asc AppServiceConfigurable] | [-sec securityTest] | [-all All]"
+	echo "$(basename "${0}"):usage: [-cd Coredata] | [-md Metadata] | [-co Command] | [-sn SupportNotification] | [-ss SupportScheduler] | [-dv DeviceVirtual] | [-asc AppServiceConfigurable] | [-sec securityTest] | [-all All]"
```
tonyespy: While shellcheck flags this line with the warning, since you're not modifying the line, and there are plenty of other shellcheck errors in this script, I'd just leave this alone.

bin/security/setupSecurityAccount.sh
```diff
@@ -2,25 +2,47 @@
...
+else
+  case ${option} in
+    -useradd)
+    echo "Info: Add security account."
+
+    OT=$(sudo edgexfoundry.security-proxy-setup-cmd --confdir=/var/snap/edgexfoundry/current/config/security-proxy-setup/res --init=false --useradd=testuser --group=admin | grep '^the access token for')
+    TOKEN=$(echo $OT | sed 's/.*: \([^.]*\.[^.]*\.[^.]*\).*/\1/')
+    export TOKEN
+
+    echo "Generated token: " $TOKEN
```
tonyespy: If there's a reason the three lines starting with `TOKEN=$(echo...` need to be different than the when `ENV_SUFFIX="-docker"` then leave them here, otherwise I suggest just using the existing three statements and moving to the after the case, so as to avoid duplicate code.

bin/setupSecurityserviceTest.sh
```diff
@@ -1,13 +1,27 @@
...
+  echo TOKEN=$TOKEN
+
+  RT=$(docker exec -i edgex-vault-worker sh -c "cat /tmp/edgex/secrets/edgex-security-proxy-setup/secrets-token.json")
+  export VAULTKEY=$(echo $RT | sed 's/.*"client_token":"\([^"]*\)".*/\1/')
+  echo VAULTKEY=$VAULTKEY
+else
+  OT=$(sudo edgexfoundry.security-proxy-setup-cmd --confdir=/var/snap/edgexfoundry/current/config/security-proxy-setup/res --init=false --useradd=jerry --group=admin | grep '^the access token for')
+  TOKEN=$(echo $OT | sed 's/.*: \([^.]*\.[^.]*\.[^.]*\).*/\1/')
+  export TOKEN
+  echo "Generated token: " $TOKEN
```
tonyespy: If there's a reason to have these statements starting with `TOKEN=$(echo...` different than the when `ENV_SUFFIX="-docker"` the leave as is, otherwise please move them after the `if/else` statement.

bin/setupSecurityserviceTest.sh
```diff
@@ -1,13 +1,27 @@
...
+
+  RT=$(sudo sh -c "cat /var/snap/edgexfoundry/current/secrets/edgex-security-proxy-setup/secrets-token.json")
+  VAULTKEY=$(echo $RT | sed 's/.*"client_token":"\([^"]*\)".*/\1/')
+  export VAULTKEY
```
tonyespy: If there's a reason to have these statements starting with `VAULTKEY=$(echo...` different than the when `ENV_SUFFIX="-docker"` the leave as is, otherwise please move them after the `if/else` statement.

deploy-edgeX-snaps.sh
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash
+
+# Set any of the following snap environment variables. Note you must also set TEST_ENV=localhost
+# (in the env.sh file) as well. Note, EF_SNAP takes precedence over EF_CHANNEL and ASC_SNAP
+# takes precedence over ASC_CHANNEL.
+
+# If EF_CHANNEL is not set or not a valid value, EF_CHANNEL will default to stable.
+# Accepted values are: stable | beta | edge | candidate
```
tonyespy: This should just say "If EF_CHANNEL is not set, it defaults to 'stable'." As you're not actually validating the value anymore, you can drop the "# Accepted values..." comment.

Also please revise this in the additional instructions section of the PR description, maybe even just point the reviewers to your snap testing README?

deploy-edgeX-snaps.sh
```diff
@@ -0,0 +1,85 @@
...
+#export EF_CHANNEL=
+
+# If ASC_CHANNEL is not set or not a valid value, ASC_CHANNEL will default to stable.
+# Accepted values are: stable | beta | edge | candidate
```
tonyespy: This should just say "If ASC_CHANNEL is not set, it defaults to 'stable'." As you're not actually validating the value anymore, you can drop the "# Accepted values..." comment.

deploy-edgeX-snaps.sh
```diff
@@ -0,0 +1,85 @@
...
+#export ASC_CHANNEL=
+
+# If EF_SNAP is not set or not a valid value, deploy-edgeX-snaps script will exit.
```
tonyespy: I think this can be dropped, as if EF_SNAP isn't set, then channel will be used, and you're not actually validating the value.

deploy-edgeX-snaps.sh
```diff
@@ -0,0 +1,85 @@
...
+# Expected absolute path of snap
+#export EF_SNAP=
+
+# If ASC_SNAP is not set or not a valid value, deploy-edgeX-snaps script will exit.
```
tonyespy: Same for ASC_SNAP

docker-compose-test-tools.yml
tonyespy: This still shows up as changed... do you know about `git checkout -- <name-of-file>`?

docs/How-to-run-blackbox-testing-against-snaps.md
```diff
@@ -0,0 +1,75 @@
...
+Uncomment the following environment variable
+
+**export TEST_ENV=localhost**
+
+If you wish to use different channels for edgexfoundry and/or app-service-configurable then navigate to the deploy-edgeX-snaps.sh file 
+and uncomment the following environment variables and append the required value 
+
+**export EF_CHANNEL=**
+
+Accepted values for EF_CHANNEL are stable, beta, edge, candidate
```
tonyespy: Not true as you can also specify tracks/branches for channels.

docs/How-to-run-blackbox-testing-against-snaps.md
```diff
@@ -0,0 +1,75 @@
...
+
+**export ASC_CHANNEL=**
+
+Accepted values for ASC_CHANNEL are stable, beta, edge, candidate
```
tonyespy: Not true as you can also specify tracks/branches for channels.

docs/How-to-run-blackbox-testing-against-snaps.md
```diff
@@ -0,0 +1,75 @@
...
+You can also use locally built snaps of either edgexfoundry and/or app-service-configurable by specifying the following environment variables
+
+**export EF_SNAP=path_to_snap**
+
+**export ASC_SNAP=path_to_snap**
+
+Replace path_to_snap with an absolute path to each snap respectively.
+
+Note, if both ASC_CHANNEL and ASC_SNAP variables are exported then ASC_SNAP will take precedence and ASC_CHANNEL will be ignored. However, if ASC_SNAP
+is invalid then ASC_CHANNEL will be used. If ASC_CHANNEL is not set i.e. exported then stable version of application-service-configurable will be installed.
```
tonyespy: Since `deploy-edgeX-snaps.sh` doesn't validate `ASC_SNAP`, this statement isn't true, the `snap install` command will fail, and the script will keep running...  Note, you can include "-e" on the shebang line of the script (e.g. `#!/bin/bash -e`) to cause the script to exist immediately when an error occurs.

bin/postman-test/collections/support-notifications-importer.postman_collection.json
```diff
@@ -1,7 +1,7 @@
 {
 	"info": {
 		"_postman_id": "068384f0-b571-47eb-841e-6417aa3b1ec7",
-		"name": "\b\bsupport-notifications-importer",
+		"name": "support-notifications-importer",
```
tonyespy, line 4: Why did you make this change?

bin/postman-test/collections/core-data-importer.postman_collection.json
```diff
@@ -1,7 +1,7 @@
 {
 	"info": {
 		"_postman_id": "f9aeb50c-4a55-4e9f-9b11-9a1b999aadf6",
-		"name": "\b\bcore-data-importer",
+		"name": "core-data-importer",
```
tonyespy, line 4: Why did you make this change?

bin/postman-test/environment/device-virtual.postman_environment.json
```diff
@@ -2,6 +2,16 @@
...
+		{
+			"key": "baseUrl",
+			"value": "http://localhost:49990",
+			"enabled": true
+		},
+		{
+			"key": "baseUrl-coreMetadata",
+			"value": "http://localhost:48081",
+			"enabled": true
+		},
```
tonyespy, line 14: I think it would make sense to make these changes in a separate commit, but I guess in the grand scheme it's probably OK to keep "as is".

bin/run.sh
```diff
@@ -11,8 +11,27 @@ fi
...
+test_with_docker() {
+  # Run the compose file for blackbox testing
+  docker-compose -f "${docker_compose_test_tools}" up -d app-service-configurable
+  export ENV_SUFFIX="-docker"
+}
+
+test_with_localhost() {
+  # Remove app-service-configurable section from docker compose file
+  sed -i '/^[[:space:]]*app-service-configurable:/,/^[[:space:]]*- edgex-app-service-configurable/d' "$docker_compose_test_tools"
+  sed -i '/^[[:space:]]*- edgex-network/d' "$docker_compose_test_tools"
```
tonyespy, line 25: Please add a comment that mentions that the script also replaces the docker network with a host network_mode.

bin/cleanSecurityserviceTest.sh
```diff
@@ -2,7 +2,11 @@
 
 echo "Info: Clean Securityservice's test data."
 
-COMPOSE=$(dirname "$0")/../$(ls $(dirname "$0")/../ | awk '/docker-compose/ && !/test-tools/')
-docker-compose -p edgex -f ${COMPOSE} run --rm --entrypoint /edgex/security-proxy-setup edgex-proxy --init=false --userdel=testuser
+if [ "$ENV_SUFFIX" = "-docker" ]; then
+  COMPOSE=$(dirname "$0")/../$(ls $(dirname "$0")/../ | awk '/docker-compose/ && !/test-tools/')
+  docker-compose -p edgex -f ${COMPOSE} run --rm --entrypoint /edgex/security-proxy-setup edgex-proxy --init=false --userdel=testuser
+else
+  sudo edgexfoundry.security-proxy-setup-cmd --init=false --userdel=testuser
```
tonyespy: When I run the tests with `SECURITY_SERVICE_NEEDED=true`, this script throws errors each time it's run:

```
Info: Delete security account.
level=ERROR ts=2020-10-23T19:34:52.001369107Z app=edgex-security-proxy-setup source=bootstrap.go:45
msg="could not load configuration file (./res/configuration.toml): open ./res/configuration.toml: no such file or directory"
```

tonyespy (changes requested):
This looks much better, but there are still some things that need to be resolved.

Also, let's chat next week about "merge master into feature" commits.

bin/postman-test/collections/support-notifications-importer.postman_collection.json
hareshkainthcanonical: At some point I updated my version with master. This change comes from https://github.com/edgexfoundry/blackbox-testing/commit/c0dda19a6d2e6bf521d8d51039bd2ef1a6e6b95b

bin/postman-test/collections/core-data-importer.postman_collection.json
hareshkainthcanonical: At some point I updated my version with master. This change comes from https://github.com/edgexfoundry/blackbox-testing/commit/c0dda19a6d2e6bf521d8d51039bd2ef1a6e6b95b

bin/setupSecurityserviceTest.sh
```diff
@@ -1,12 +1,22 @@
...
+  COMPOSE=$(dirname "$0")/../$(ls $(dirname "$0")/../ | awk '/docker-compose/ && !/test-tools/')
+  OT=$(docker-compose -p edgex -f ${COMPOSE} run --rm --entrypoint /edgex/security-proxy-setup edgex-proxy --init=false --useradd=jerry --group=admin | grep '^the access token for')
+  RT=$(docker exec -i edgex-vault-worker sh -c "cat /tmp/edgex/secrets/edgex-security-proxy-setup/secrets-token.json")
+
+else
+  OT=$(sudo edgexfoundry.security-proxy-setup-cmd --confdir=/var/snap/edgexfoundry/current/config/security-proxy-setup/res --init=false --useradd=jerry --group=admin | grep '^the access token for')
+  RT=$(sudo sh -c "cat /var/snap/edgexfoundry/current/secrets/edgex-security-proxy-setup/secrets-token.json")
+fi
+
+TOKEN=$(echo $OT | sed 's/.*: \([^.]*\.[^.]*\.[^.]*\).*/\1/')
```
tonyespy, line 14: I would've kept the export and the set as a single statement, but this doesn't warrant nak'ing the PR. Next time, remember that you always want to change as little as possible...

bin/setupSecurityserviceTest.sh
```diff
@@ -1,12 +1,22 @@
...
+export TOKEN
 echo TOKEN=$TOKEN
 
-RT=$(docker exec -i edgex-vault-worker sh -c "cat /tmp/edgex/secrets/edgex-security-proxy-setup/secrets-token.json")
-export VAULTKEY=$(echo $RT | sed 's/.*"client_token":"\([^"]*\)".*/\1/')
+VAULTKEY=$(echo $RT | sed 's/.*"client_token":"\([^"]*\)".*/\1/')
+export VAULTKEY
```
tonyespy, line 19: Same as above...

deploy-edgeX-snaps.sh
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash
+
+# Set any of the following snap environment variables. Note you must also set TEST_ENV=localhost
+# (in the env.sh file) as well. Note, EF_SNAP takes precedence over EF_CHANNEL and ASC_SNAP
+# takes precedence over ASC_CHANNEL.
+
+# If EF_CHANNEL is not set, it defaults to 'stable'.
+# Accepted values are: stable | beta | edge | candidate | channel/track e.g. latest/edge
```
tonyespy: I would just drop "Accepted values..." altogether.  "channel/track e.g. latest/edge" doesn't really make sense. A channel is a triplet of "track" + "risk" + "branch", so "latest/edge" is really track/risk, not a "channel/track". In the end, someone who knows snaps will understand how to set this.

deploy-edgeX-snaps.sh
```diff
@@ -0,0 +1,85 @@
...
+#export EF_CHANNEL=
+
+# If ASC_CHANNEL is not set, it defaults to 'stable'.
+# Accepted values are: stable | beta | edge | candidate | channel/track e.g. latest/edge
```
tonyespy: Same as above...

tonyespy (changes requested):
Super close... a couple of minor comments inline, and you also need to fix your commit message so that the DCO check passes.

tonyespy (approved):
LGTM

hareshkainthcanonical:
Hi @jinlinGuan  could you please review the PR?

jinlinGuan (approved):
LGTM

cloudxxx8 (approved):
approve based on Tony and Ginny's review

